PR #1, opened 2018-08-14
new landing page

bianca:
Preliminary new landing page, including new snippets for the dataset stat, introduction, featured datasets, search, and contact info


ckanext/ontario_theme/fanstatic/ontario_theme.less
```diff
@@ -1,5 +1,25 @@
 @import "smarties.less";
```
boykoc: Looks like the compiled less files into css were missed. Can you add this and resubmit?

PR #13, opened 2018-10-16
Feature: Match Ontario.ca style

kalcevka:
Refinements to the Colby look and feel to more closely match with Ontario.ca styling. Rewrites to some home page text for clarity/brevity. Restyle some home page elements so that there are fewer call-to-actions and only a few key elements are highlighted. Revise heading hierarchy for improved accessibility.


boykoc:
Closing as this is replaced by PR #17.

PR #37, opened 2019-02-15
Legacy ui improvements

bianca:
UI adjustments stemming from first 6 user tests. Changes include:
 - apply the grey rounded corner style of the filters on the dataset search page to the top search box on the dataset search page
 - change the styles to the dataset manage button to draw attention to it and change it to "edit"
 - Change the "Next: add resources" button in add dataset form to "Save and add resources" to reassure users that their metadata is saved
 - make sure tags in view dataset page don't truncate
 - have each organization in view organizations page span the whole page
 - break out each "explore" menu for each resource
 - search form in dataset search page changed to span the whole page


boykoc:
I'm not sure about some of the changes but we can amend them in the future if needed.

PR #38, opened 2019-02-19
Mystery resource filesize

bianca:
This branch contains a single change, removing "(unknown size)" from each resource item. This programmatically appears when there is no file size for a resource, but will practically mostly appear for remote resources, but user feedback deemed it confusing.


boykoc:
Resubmitting clean against master.

PR #39, opened 2019-02-19
More filetype icons

bianca:
This branch consists of two changes, the additions of .shp and .kml resource icons.


boykoc:
Resubmitting clean against master.

PR #42, opened 2019-03-15
Improvement: remove '(unknown size)' from resources

bianca:
This stems from a request from a current Colby user to hide "(unknown size)" next to resources that are links and therefore wouldn't have a filesize anyways. While this solution hides all resources that have no filesize (not just resources that are links), it effectively solves the issue.


boykoc:
Just clarifying,

> "While this solution hides all resources that have no filesize (not just resources that are links), it effectively solves the issue."

Is referring to hiding the text of "(unknown size)" for all resources that have no file size. Not that it hides all resources with no file size.

PR #46, opened 2019-03-20
Harmonize search bar look

bianca:
This pull request consists of one commit to address:
https://trello.com/c/QN1a5T9f/385-harmonizing-the-look-of-the-search-bar-in-the-datasets-search-page-with-the-left-side-filtering-options-matching-with-the-grey-o

It changes the organizations search bar to match the packages search bar.

### DIFF
--- a/ckanext/ontario_theme/templates/snippets/add_dataset.html
+++ b/ckanext/ontario_theme/templates/snippets/add_dataset.html
@@ -1,0 +1,10 @@
+{# Adds 'Add Dataset' button #}
+
+{% set dataset_type = dataset_type if dataset_type else 'dataset' %}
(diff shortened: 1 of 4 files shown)


ckanext/ontario_theme/templates/snippets/add_dataset.html
```diff
@@ -0,0 +1,10 @@
+{# Adds 'Add Dataset' button #}
```
boykoc: Unsure where this file is used within this PR and set of changes.

boykoc: just wanted to note that. I'm accepting for now and we can undo changes in future.

PR #47, opened 2019-03-20
Improvement: Language toggle link added in header for english and french

bianca:
This pull request has one major change to address:
https://trello.com/c/yoZPZRE8/388-language-selector-missing-from-site

A EN/FR toggle was added to the header.


boykoc:
Issue with jenkins causing checks to fail.

PR #57, opened 2019-06-18
Improvement: Datapusher template override for user friendly error message.

boykoc:
Error message was giving much more detailed information than
needed to the user. This prevents divulging unnecessary information
in the web UI.


boykoc:
The theme extension must be before the datapusher extension for this to work as expected.

PR #66, opened 2019-08-15
French content for public data catalogue

bianca:
This pull request enables the inclusion of french translation in the ontario_theme. 

It consists of:
- changing the templates to include {% trans %} and {% endtrans %} tags so that the translation template can be generated
- generating the general translation template
- generating the french-specific po file for populating french translations
- updating the po file with french translations
- including the mo file (compiled from po file)
- updating plugin.py to enable the use of french translations


ckanext/ontario_theme/templates/home/help.html
```diff
@@ -10,282 +10,282 @@
   <article class="module">
     <div class="module-content">
       <h1 class="page-heading">{{ _('Help') }}</h1>
-      <p>Need more help or want to learn more about the data catalogue? <a href="mailto:opengov@ontario.ca">Contact the Open Government team</a> — we’ll gladly answer any questions you might have!</p>
+
```
boykoc: As discussed, ideally this would be one large translation but we'll leave as is for now as it creates issues when generating the msgid's. 

[example where it's manually done and nicely formatted](https://github.com/open-data/ckanext-canada/blob/2bb32483b47c2a91a2717eefb2984d65d5d0a518/ckanext/canada/i18n/fr/LC_MESSAGES/canada.po#L1873). Bonus, more control and can keep it cleaner. Downside, it doesn't give those nice line numbers and file references.

PR #69, opened 2019-08-19
About content (both french and english)

bianca:
This pull request consists of 7 commits that add about content to the site, in both french and english. The changes consist of:

- Adding the about_text snippet to the codebase
- Adding the english content to the about_text snippet
- Changes to card.less to be able to apply it to non-anchored cards (for the metadata section of the content)
- French translations for the content
- Content fixes


bianca:
correction: 8 commits

PR #87, opened 2019-10-28
Task: Add public catalogue id field to Colby

bianca:
This PR consists of one change and one commit: In order to link the colby and public catalogue records together, a field to store the public catalogue id needs to be added to colby.


boykoc:
Closing for time being.

PR #89, opened 2019-11-04
Task: Add resource details in view package page

bianca:
This PR consists of one change in one commit - Resource language and date range was added to each resource in the resource list section of the view package page.


boykoc:
Out of date, replaced by #99

PR #91, opened 2019-11-07
Fluent organizations

bianca:
This PR implements multilingual organization titles and descriptions. the PR consists of 3 commits:
- addition of the organizations schema (Please do the following to the ini file:
 1. add scheming_organizations' to ckan.plugins
 2. add 'scheming.organization_schemas = ckanext.ontario_theme:ontario_theme_organization.json' to the scheming part of the ini file)
- addition of snippets and blocks containing organization display_name and/or description to the codebase
- updating snippets to use the translated title and description fields

Please note that the change to the new fields is almost, but not quite, global. One change still to be done is converting display_name to title_translated in "My Organizations" when someone is logged in and using their dashboard. The results of the helper organizations_available is passed to organizations_list, but for some reason does not contain the extra fields. This fix is to-be-done.


boykoc:
Needs more testing / fixing.

PR #96, opened 2019-12-03
Odc beta add language disclaimer

bianca:
This PR consists of 3 commits to make 1 change: the addition of a language disclaimer on the resource page that displays in 4 cases:

- you're viewing the site in french and looking at an english only resource. there are french resources available.
- you're viewing the site in french and looking at an english only resource. there are no french resources available.
- you're viewing the site in english and looking at a french only resource. there are english resources available.
- you're viewing the site in english and looking at a french only resource. there are no english resources available.

### DIFF
--- a/ckanext/ontario_theme/templates/package/snippets/ontario_theme_language_disclaimer.html
+++ b/ckanext/ontario_theme/templates/package/snippets/ontario_theme_language_disclaimer.html
@@ -1,0 +1,23 @@
+{#
+  Custom language notification in the cases that users are viewing the site in one language and a resource in another
+#}
+{% set current_lang = request.environ.CKAN_LANG %}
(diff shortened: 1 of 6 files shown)


ckanext/ontario_theme/templates/package/snippets/ontario_theme_language_disclaimer.html
```diff
@@ -0,0 +1,23 @@
+{#
+  Custom language notification in the cases that users are viewing the site in one language and a resource in another
+#}
+{% set current_lang = request.environ.CKAN_LANG %}
+{% if (res.language == "english" and current_lang == "fr") or (res.language == "french" and current_lang == "en") %}
```
boykoc: For future time, I think it would be good to walkthrough these conditionals. Essentially wondering if they are needed or add an extra layer of complexity for us and the user.  I had imagined these would just always show (e.g. "this resource is in [language x] only.").

PR #103, opened 2019-12-27
french licences

bianca:
This PR consists of one commit for changes to allow for both french and english licences. 

Changes include:
- adding french text to the licences json file.
- adding a helper to get the translated title
- having licence info toggle between english and french on the dataset page
- changing the licence snippet on the dataset page to only be hyperlinked if there is a url.
- having the licence name toggle between english and french in the facet list on the dataset search results page 

Note: This PR uses the core get_translated function, which wasn't intended for licences.


bianca:
These changes were made on a branch forked off of odc-beta but they should have been forked off of master. See PR #107

PR #112, opened 2020-01-10
Task: Add CSRF for Pylons and Flask.

boykoc:
CKAN is currently running both pylons and flask and migrating
towards only flask. Current CSRF approaches only address pylons
requests.

After trying numerous options this seemed like the most reliable and enables
easy throw away once pylons is completely gone from CKAN.

Basic concept is to run the original CSRF with very minor changes. Then use a
modified version to handle Flask.

I used 2 repos as my bases: [ckanext-security](https://github.com/data-govt-nz/ckanext-security) and [ckan-ex-qgov](https://github.com/qld-gov-au/ckan-ex-qgov).

anti_csrf3.py is the modified [middleware.py](https://github.com/data-govt-nz/ckanext-security/blob/master/ckanext/security/middleware.py) that breaks apart the classes
and modified for Flask. I liked most of how this version, ckanext-security,
was implemented which is why I used it instead of the ckan-ex-qgov.

ckan-ex-qgov I used as the base for pylons as it didn't
require middleware or modifying CKAN directly (middleware implementation).

**File references:**

1. orig_anti_csrf.py was from https://github.com/qld-gov-au/ckan-ex-qgov/blob/master/ckanext/qgov/common/anti_csrf.py
2. anti_csrf.py was from https://github.com/data-govt-nz/ckanext-security/blob/master/ckanext/security/anti_csrf.py
3. anti_csrf3.py was from https://github.com/data-govt-nz/ckanext-security/blob/master/ckanext/security/middleware.py

### DIFF
--- a/ckanext/ontario_theme/anti_csrf.py
+++ b/ckanext/ontario_theme/anti_csrf.py
@@ -1,0 +1,283 @@
+# encoding: utf-8
+"""Provides a self-contained filter to prevent Cross-Site Request Forgery,
+based on the Double Submit Cookie pattern,
+www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Double_Submit_Cookie
+The filter can be enabled simply by invoking 'intercept_csrf()'.
+"""
+import hashlib
+import hmac
+import random
+import re
+from re import IGNORECASE, MULTILINE
+import time
+import urllib
+from logging import getLogger
+
+from ckan.common import config, request, response, g
+import ckan.lib.base as base
+
+LOG = getLogger(__name__)
+
+RAW_RENDER_JINJA = base.render_jinja2
+RAW_BEFORE = base.BaseController.__before__
+
+""" Used as the cookie name and input field name.
+"""
+TOKEN_FIELD_NAME = 'token'
+
+"""
+This will match a POST form that has whitespace after the opening tag (which all existing forms do).
+Once we have injected a token immediately after the opening tag,
+it won't match any more, which avoids redundant injection.
+"""
+POST_FORM = re.compile(r'(<form [^>]*method=["\']post["\'][^>]*>)(\s[^<]*<)', IGNORECASE|MULTILINE)
+
+"""The format of the token HTML field.
+"""
+TOKEN_VALIDATION_PATTERN = re.compile(r'^[0-9a-z]+![0-9]+/[0-9]+/[-_a-z0-9%]+$', IGNORECASE)
+API_URL = re.compile(r'^/api\b.*')
+CONFIRM_MODULE_PATTERN = r'data-module=["\']confirm-action["\']'
+HREF_URL_PATTERN = r'href=["\']([^"\']+)'
+
+# We need to edit confirm-action links, which get intercepted by JavaScript,
+#regardless of which order their 'data-module' and 'href' attributes appear.
+CONFIRM_LINK = re.compile(r'(<a [^>]*{}[^>]*{})(["\'])'.format(
+    CONFIRM_MODULE_PATTERN, HREF_URL_PATTERN),
+                          IGNORECASE|MULTILINE)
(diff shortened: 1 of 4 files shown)


ckanext/ontario_theme/anti_csrf.py
```diff
@@ -0,0 +1,143 @@
+"""Provides a filter to prevent Cross-Site Request Forgery,
```
ThrawnCA: Is this file really necessary when we have `orig_anti_csrf.py`?

ckanext/ontario_theme/anti_csrf3.py
```diff
@@ -0,0 +1,121 @@
...
+from flask import Flask, make_response, render_template, request
+from ckan.common import config, is_flask_request
+import ckan.lib.base as base
+
+import logging
+
+log = logging.getLogger(__name__)
+
+
+CSRF_ERR = 'CSRF authentication failed. Token missing or invalid.'
```
ThrawnCA: This isn't used?

ckanext/ontario_theme/anti_csrf3.py
```diff
@@ -0,0 +1,121 @@
...
+#        return is_secure() and good_referer() and \
+#               good_origin() and check_token()
+    return check_token()
+
+##############################################################################
+
+
+def is_secure():
+    # allow requests which have the x-forwarded-proto of https (inserted by nginx)
+    if request.headers.get('X-Forwarded-Proto') == 'https':
```
ThrawnCA: Headers can be forged.

ckanext/ontario_theme/anti_csrf3.py
```diff
@@ -0,0 +1,121 @@
...
+    return request.scheme == 'https'
+
+def is_safe():
+    "Check if the request is 'safe', if the request is safe it will not be checked for csrf"
+    # api requests are exempt from csrf checks
+    if request.path.startswith("/api"):
+        return True
+
+    # get/head/options/trace are exempt from csrf checks
+    return request.method in ('GET', 'HEAD', 'OPTIONS', 'TRACE')
```
ThrawnCA: TRACE should probably be entirely blocked, not given special treatment. It actually enables some types of CSRF filter bypass, by exposing the token cookie.

ckanext/ontario_theme/anti_csrf3.py
```diff
@@ -0,0 +1,121 @@
...
+    if not request.method == 'POST' and (request.args.get(anti_csrf.TOKEN_FIELD_NAME) and len(request.args.get(anti_csrf.TOKEN_FIELD_NAME)) == 1):
+        token = request.args.get(TOKEN_FIELD_NAME)
+        #del request.GET[anti_csrf.TOKEN_FIELD_NAME]
+        return token
+    post_tokens = request.form.getlist(anti_csrf.TOKEN_FIELD_NAME)
+    if not post_tokens or len(post_tokens) != 1:
+        return None
+    token = post_tokens[0]
+    # drop token from request so it doesn't populate resource extras
+    #del request.POST[anti_csrf.TOKEN_FIELD_NAME]
```
ThrawnCA: Why is this commented out? Does it not apply to Flask requests?

ckanext/ontario_theme/anti_csrf3.py
```diff
@@ -0,0 +1,121 @@
...
+import ckan.lib.base as base
+
+import logging
+
+log = logging.getLogger(__name__)
+
+
+CSRF_ERR = 'CSRF authentication failed. Token missing or invalid.'
+
+domain = config.get('ckan.ontario_theme.csrf_domain', '')
```
ThrawnCA: Referrer checks should probably be optional, but if they're going to be implemented, shouldn't they be derived from `ckan.site_url`?

ThrawnCA (changes requested):

ThrawnCA:
@boykoc This is implemented in https://github.com/qld-gov-au/ckanext-csrf-filter/tree/develop as a stand-alone extension now. Comments there are welcome.

PR #117, opened 2020-04-02
Task: Add link to Open Data Guidebook for exemption field.

bianca:
This PR includes one commit to make two changes:

 * add a link to the open data guidebook
 * enable markdown and html for help text in the form so that the link can be opened in a new window 

In order to provide more guidance in the add dataset form, for the exemption field a link to the open data guidebook was added. In order to enable this so that the link could be opened in a new window (since the link is in a form), allow_html is enabled for help text.

### DIFF
--- a/ckanext/ontario_theme/templates/scheming/form_snippets/help_text.html
+++ b/ckanext/ontario_theme/templates/scheming/form_snippets/help_text.html
@@ -1,0 +1,8 @@
+{% import 'macros/form.html' as form %}
+
+{%- if field.help_text -%}
+    {{- form.info(
+        text=h.render_markdown(h.scheming_language_text(field.help_text), True, True),
+        inline=field.get('help_inline', false)
+        ) -}}
+{%- endif -%}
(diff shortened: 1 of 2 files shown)


ckanext/ontario_theme/templates/scheming/form_snippets/help_text.html
```diff
@@ -0,0 +1,8 @@
+{% import 'macros/form.html' as form %}
+
+{%- if field.help_text -%}
+    {{- form.info(
+        text=h.render_markdown(h.scheming_language_text(field.help_text), True, True),
```
boykoc, line 5: Seems fine as this snippet is not used with user entered values.

boykoc, line 5: For future, could call the `render_markdown()` helper with named args like this for readability.
`h.render_markdown(h.scheming_language_text(field.help_text), allow_html=True)`

PR #131, opened 2020-04-27
Dependent field validator

bianca:
This PR consists of two features, both applied to the exemption and exemption rationale fields:
 * a ckan js module that shows and hides a field based on the value of another field
 * a form validator that removes the values of a field unless another field has a certain value


bianca:
This PR was made on an unfinished feature. The validator does not work.

PR #132, opened 2020-05-22
Make access level required

bianca:
Access level is a required piece of information for the public data catalogue.

This PR consists of one change (3 commits) to make access level required. Changing "required" to "true" is not sufficient to assure the right form behaviour, the scheming_required validator must be added. The "form_include_blank_choice" flag was set to "true" will make sure the user doesn't just unknowningly go with a set default value for access_level. 

This establishes preferred behaviour: If the user doesn't see/fill out access level, the validator sends them back.


boykoc:
making it required with an empty choice does NOT allow the empty choice.  So although some validators /settings sound weird, it's correct.

PR #133, opened 2020-05-22
Change new dataset disclaimer

bianca:
The new dataset disclaimer wasn't relevant to the public catalogue.

This PR consists of 1 commit to change the disclaimer to be public catalogue-specific.

Note: when more guidance is available, it can be added.

### DIFF
--- a/ckanext/ontario_theme/templates/package/snippets/disclaimer.html
+++ b/ckanext/ontario_theme/templates/package/snippets/disclaimer.html
@@ -8,8 +8,7 @@
   <p>
     <i class="fa fa-info-circle" 
        aria-hidden="true"></i> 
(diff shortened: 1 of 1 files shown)


ckanext/ontario_theme/templates/package/snippets/disclaimer.html
```diff
@@ -8,8 +8,7 @@
   <p>
     <i class="fa fa-info-circle" 
        aria-hidden="true"></i> 
-    By adding or editing this you are confirming that this data / information 
-    is appropriate to share with the OPS.
+    By adding or editing this you are confirming that this data and/or metadata
```
boykoc, line 11: Missing translation, but seems it was before too.

PR #135, opened 2020-06-19
Task: Consolidate public and internal branches.

boykoc:
### What the Change Is

Currently, the internal/base catalogue code is in the master branch and the external catalogue changes/overrides are in the odc-beta branch.

This PR merges the odc-beta branch back into master and keeps external and internal/base code separated through different plugins and template directories instead of duplicate and diverging branches.

A comparison was done of odc-beta against master to determine the differences. In the end, there were actually very few differences that were necessary to keep. Many things in odc-beta could be applied to the base templates with only 14 template files kept as overrides as well as other differences in assets, schemas and plugin classes.

### Why Make the Change

This repo originally started as an internal data catalogue project. Then it was modified to be also used for a new external data catalogue. At the start it was a little unclear how exactly the 2 systems would interact and how similar (or different) the designs/templates would need to be. There was a desire to keep it as a single repository as there were many similar portions and mostly the styles and language in templates were different.

The original idea was to push common changes into master and use this internally. Anything external specific would be made against the odc-beta branch. Master would be merged into odc-beta to apply any common changes. However, this added complexity to the developer workflow:

- developers had to determine the best spot to branch from to start work on a task (i.e. from master or from odc-beta)
- the 2 branches started to become separate repo's with a lot of duplicate code
- issues handling internal only changes that started to come up, essentially had to apply to master, merge to odc-beta, then do seperate PR against odc-beta undoing changes
- started to become a bottle-neck doing code-reviews and merge conflicts and error prone

#### Considerations:

- break out to 2 repos
  - Pros: somewhat different projects so keeps it clean, single focus, clean commits and branching structure.
  - Cons: duplication of code, duplicate commits for common items, 2 separate code-bases to maintain and update.
- single repo with some form of branching or gitflow workflow to try and address the issue
  - Pros: single code-base, helps encourage similarity where possible/helps prevent unnecessary divergence, 
  - Cons: kind of what we're currently doing and its caused us to look at alternatives, confusing to know which branch to work from/where to put commits, complicated git workflow for small project, someone needing to manage merge conflicts and look for common changes to apply to all branches, had to run one site from non-master/non-release branch
- **[Implemented in this PR]** single repo with code structure changes (e.g. 2 plugins, separate internal/external directories/files as needed)
  - Pros: single code-base,  helps encourage similarity where possible/helps prevent unnecessary divergence, uses CKAN's plugin and template override practices, enables common place to put code/templates as needed, clean spot to put different code/templates, examples of this in the wild, cleaner git workflow (i.e. always branch from master)
  - Cons: CKAN's plugin loading order and layers of template overrides can get confusing and can cause issues, ties 2 catalogues closer together as it's using same branch.


ckanext/ontario_theme/templates/package/read_base.html
```diff
@@ -1,10 +0,0 @@
-{% ckan_extends %}
```
boykoc, line 1: replaced above.

ckanext/ontario_theme/templates/package/resource_read.html
```diff
@@ -1,42 +0,0 @@
-{% ckan_extends %}
```
boykoc, line 1: replaced above.

ckanext/ontario_theme/templates/scheming/package/read.html
```diff
@@ -1,74 +0,0 @@
-{% extends "package/read.html" %}
```
boykoc, line 1: replaced above.

PR #138, opened 2020-07-08
Groups to fluent collections

bianca:
This PR consists of 8 commits to make the following changes:

- Implement groups in scheming
- Implement fluent/bilingual groups
- Update templates for fluent groups
- Add scheming group form formatting
- Change group type for the site from "group" to "collection"
- Change templates to use "collection" label for groups
- Update translations to use "collection" for groups
- Add routes to assure consistent /collections pattern 

Add scheming_groups to ckan.plugins in the ini file when including this PR.


ckanext/ontario_theme/templates/internal/group/index.html
```diff
@@ -1,19 +1,23 @@
 {% ckan_extends %}
 
+{% block breadcrumb_content %}
+  <li class="active">{% link_for _('Collections'), controller='group', action='index', named_route=group_type + '_index' %}</li>
```
boykoc, line 4: Instead of hardcoding the collection type name we could switch this to something like

`<li class="active">{% link_for _(group_type.title()), controller='group', action='index', named_route=group_type + '_index' %}</li>`

This allows us to make one change and have the terms (i.e. "Collection", "Group", "New thing") updated automatically.

This should apply to most template updates made in this PR and hopefully help prevent changes for future us.

ckanext/ontario_theme/templates/internal/header.html
```diff
@@ -87,7 +87,7 @@ <h1>
         {{ h.build_nav_main(
           ('search', _('Datasets')),
           ('organizations_index', _('Organizations')),
-          ('group_index', _('Groups')),          
+          ('group_index', _('Collections')),
```
boykoc, line 90: This is one area we probably do want to hard code the change for so that our navigation works as we want it. But we should change the route to collections.

`('collection_index', _('Collections')),`

ckanext/ontario_theme/templates/internal/group/snippets/helper.html
```diff
@@ -0,0 +1,16 @@
...
+    <p>
+      {% trans %}
+        You can create and manage collections of datasets.
+        This could be to catalogue datasets for a particular project or team,
+        or on a particular theme, or as a very simple way to help people find
+        and search your own published datasets.
+      {% endtrans %}
+    </p>
+  </div>
+</div>
```
boykoc, line 16: Could be it's own helper.html snippet for collections and load the right snippet based on collection type. May be overkill for this one but allows a bit more extensibility in the future. This may be more applicable to various dataset types if/when we get to those.

ckanext/ontario_theme/templates/internal/header.html
bianca: agreed. change made.

ckanext/ontario_theme/templates/internal/group/index.html
bianca: It was agreed to make all these changes to ckanext-scheming core instead.

boykoc:
Out of date - replaced by PR #160

PR #143, opened 2020-08-12
Add french ga tags

bianca:
This PR consists of 2 commits to:
 - add a get_translated function that takes a specified language
 - use that function to create ministry-name-fr and dataset-name-fr attributes for GA use on organization and dataset pages

### DIFF
--- a/ckanext/ontario_theme/plugin.py
+++ b/ckanext/ontario_theme/plugin.py
@@ -9,6 +9,7 @@ from flask import render_template, render_template_string
 import ckanapi_exporter.exporter as exporter
 import json
 import ckan.lib.i18n as i18n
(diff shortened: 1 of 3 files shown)


ckanext/ontario_theme/plugin.py
```diff
@@ -224,6 +226,13 @@ def get_license(license_id):
     '''
     return Package.get_license_register().get(license_id)
 
+def get_translated_lang(data_dict, field, specified_language):
+    site_language = i18n.get_lang()
```
boykoc, line 230: This doesn't seem to be needed. Could also drop line 11 above as well.

boykoc (approved):

PR #150, opened 2020-09-03
Accessibility repair

bianca:
This PR consists of 5 commits to make the following changes:
* update colours to accessible colours from the ontario.ca design system
* shift to use of bootstrap's sr-only class for screen-reader-only text
* addition of aria-label and aria-hidden to some buttons that use icons
* update homepage with correct header order and consistent colour contrasting

To do in a future PR:
- headings order needs to be reviewed on other pages
- potentially review all icon buttons in order to add accessible labels

### DIFF
--- a/ckanext/ontario_theme/templates/internal/header.html
+++ b/ckanext/ontario_theme/templates/internal/header.html
@@ -88,9 +88,9 @@
   {% endif %} {% endblock %}
   <div class="container">
     <div class="navbar-right">
-      <button data-target="#main-navigation-toggle" data-toggle="collapse" class="navbar-toggle collapsed" aria-expanded="false" type="button">
-        <span class="sr-only">Toggle navigation</span>
-        <span class="fa fa-bars"></span>
+      <button data-target="#main-navigation-toggle" data-toggle="collapse" class="navbar-toggle collapsed" aria-expanded="false" type="button" aria-label="{{ _('Toggle navigation') }}">
+        <span class="sr-only">{{ _('Toggle navigation') }}</span>
+        <span class="fa fa-bars" aria-hidden="true"></span>
       </button>
     </div>
     <hgroup class="{{ g.header_class }} navbar-left">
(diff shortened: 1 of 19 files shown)


ckanext/ontario_theme/templates/internal/header.html
```diff
@@ -127,7 +127,7 @@
           <div class="field">
             <label for="field-sitewide-search">{% block header_site_search_label %}{{ _('Search Datasets') }}{% endblock %}</label>
             <input id="field-sitewide-search" type="text" class="form-control" name="q" placeholder="{{ _('Search') }}" />
-            <button class="btn-search" type="submit"><i class="fa fa-search"></i><span>Submit Search</span></button>
+            <button class="btn-search" type="submit" aria-label="{{ _('Submit Search') }}"><i class="fa fa-search" aria-hidden="true"></i><span class="sr-only">{{ _('Submit Search') }}</span></button>
```
boykoc, line 130: Caution on this. aria-hidden true removes this element and child elements for assistive technology. Adding a child element with sr-only implies it's intended for screen readers.

ckanext/ontario_theme/templates/internal/home/snippets/search.html
```diff
@@ -10,8 +10,8 @@
   <form class="module-content search-form" method="get" action="{% url_for controller='package', action='search' %}">
     <div class="search-input form-group search-giant">
       <input aria-label="{% block header_site_search_label %}{{ _('Search datasets') }}{% endblock %}" id="field-main-search" type="text" class="form-control" name="q" value="" autocomplete="off" placeholder="{% block search_placeholder %}{{ placeholder }}{% endblock %}" />
-      <button type="submit">
-        <i class="fa fa-search"></i>
+      <button type="submit" aria-label="{{ _('Search') }}">
```
boykoc, line 13: This is a bit duplicative but manual testing with assistive technology would help see a bit more. adding the aria-label while using an inner element as the descriptive text may cause "search" to be read / identified twice.

ckanext/ontario_theme/templates/internal/snippets/search_form.html
```diff
@@ -1,5 +1,15 @@
 {% ckan_extends %}
 
+{% block search_input_button %}
+<span class="input-group-btn">
+  <button class="btn btn-default btn-lg" type="submit" value="search" aria-label="{{ _('Search') }}">
```
boykoc, line 5: Similar to search.html and menu in header.html

boykoc (approved):

PR #161, opened 2020-10-21
Fix organization display name

bianca:
This PR consists of 2 commits to fix cases where display_name is still being used to display an organization's name (instead of get_translated).


boykoc (approved):
LGTM

PR #178, opened 2021-03-11
Task: Modified csv_dump to produce ordered columns at /dataset/inventory

hq-ods:
- changed to store columns in an ordered dictionary due to python 2 default behavior of storing dictionary items in unordered pairs
- changed to use utf-8 with BOM due to excel display issues
- updated file name and include a datastamp
- changed url to /dataset/inventory instead of /dataset/csv_dump

### DIFF
--- a/ckanext/ontario_theme/plugin.py
+++ b/ckanext/ontario_theme/plugin.py
@@ -150,19 +150,6 @@ def csv_dump():
         }),
         ("Keywords FR", {
             "pattern": ["^keywords$", "^fr$"]
-        }),
-        ("Contains geopgrapic markers", {
-            "pattern": ["^resources$", "^contains_geographic_markers$"],
-            "deduplicate": "true"
-        }),
-        ("License Id", {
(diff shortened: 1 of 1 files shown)


ckanext/ontario_theme/plugin.py
```diff
@@ -78,176 +80,101 @@ def csv_dump():
...
+        ("Geographic Coverage FR", {
+            "pattern": ["^geographic_coverage_translated$", "^fr$"]
+        }),
+        ("Exemption Rationale FR", {
+            "pattern": ["^exemption_rationale$", "^fr$"]
+        }),
+        ("Keywords FR", {
+            "pattern": ["^keywords$", "^fr$"]
+        }),
+        ("Contains geopgrapic markers", {
```
bianca, line 154: Can we ditch "contains geographic markers", "data range start", "data range end", and "licence id"?

bianca (changes requested):

PR #227, opened 2022-08-02
update version and contact info in setup.py

KatiRG:
## What this PR accomplishes
Updates [setup.py](https://github.com/ongov/ckanext-ontario_theme/blob/ckan_2.9_ver/setup.py) to include:
- correct python version (3.8)
- current contact info

Updates CKAN version to 2.9.x in [README](https://github.com/ongov/ckanext-ontario_theme/blob/ckan_2.9_ver/README.rst)

## What needs review
- confirm that contact info is correct

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ setup(
     long_description=long_description,
 
     # The project's main homepage.
-    url='https://github.com/boykoc/ckanext-ontario_theme',
+    url='https://github.com/ongov/ckanext-ontario_theme',
 
     # Author details
     author='''Ontario Digital Service''',
--- a/setup.py
+++ b/setup.py
@@ -26,8 +26,8 @@ setup(
     url='https://github.com/boykoc/ckanext-ontario_theme',
 
     # Author details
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -26,8 +26,8 @@
     url='https://github.com/boykoc/ckanext-ontario_theme',
```
aimalkhan, line 26: this should be updated to `https://github.com/ongov/ckanext-ontario_theme`

aimalkhan, line 26: `version='0.0.1'` on line 20 should be corrected to the versioning scheme we are using. I would like a comment on this from @bianca 
current release v1.0 according to tags (https://github.com/ongov/ckanext-ontario_theme/releases)

aimalkhan (commented):
Looks good, but i wanted a couple of changes. I want the requested theme version changes to be reviewed by @bianca

setup.py
KatiRG: sorry! missed that

aimalkhan: update to 2.0.0, please.

Issue #237, opened 2022-10-04
Configure Google Tag Manager to track different views


KatiRG:
Set up a dataLayer snippet for event `Pageview` to load on every page view, passing variables for ministry name, group name, availability status of dataset, dataset name, and resource name. Where not applicable (i.e. for views other than dataset page views and resource page views), pass in blank strings for these variables.

- [x] ckan 2.8
- [x] ckan 2.9

KatiRG:
[Known issue](https://www.analyticsmania.com/post/multiple-installations-of-google-tag-manager-detected/) with legacy Google Tag Assistant:

The assistant may report "multiple installations of google tag manager detected", but to determine if this is the case, check how many times `gtm.js` is loaded in in the browser Network tab.

PR #238, opened 2022-10-04
Google Tag Manager: track page and resource views

KatiRG:
## What this PR accomplishes
Sets up a dataLayer snippet in [gtm_pageview.html](https://github.com/ongov/ckanext-ontario_theme/blob/gtm/ckanext/ontario_theme/templates/internal/gtm_pageview.html) with a `Pageview` event that fires on every page and contains placeholders for the following variables:

```
'ministryTag'
'group'
'dataAvailability'
'datasetName'
'resourceName'
```
Values for these variables will be passed by dataset view pages and resource view pages; all other pages will pass empty strings. 

Conditional logic is used in [base.html](https://github.com/ongov/ckanext-ontario_theme/blob/ckan_2.9_gtm/ckanext/ontario_theme/templates/internal/base.html) to assign package-related and resource-related values when `pkg` or `res` objects exist; the desired values are then extracted from these objects in  [gtm_pageview.html](https://github.com/ongov/ckanext-ontario_theme/blob/gtm/ckanext/ontario_theme/templates/internal/gtm_pageview.html).

## Issue addressed
#237, ckan 2.8

## What needs review
- verify with Google Tag Manager Dashboard that `Pageview` event is working as expected
- check in the Network tab of the browser that `gtm.js` is loaded only once on each page load (see discussion in #237).

Caveat: 
At the moment, all pages are loaded twice when clicking on any resources, causing the `dataLayer` snippet to send two `Pageview` events. This issue is unrelated to Google Tags and will be addressed separately.


aimalkhan (approved):
👍

Issue #243, opened 2022-10-19
Track resource download for Google analytics


KatiRG:
Set up a dataLayer for event "File Downloads" that triggers every time a download button or download URL is clicked. Download buttons are found in dataset pages and resource pages. Download URLs are only found on resource pages. The dataLayer should be of the form:
```
window.dataLayer.push({
    'event': 'File Download',
    'ministryTag': ,
    'group': ,
    'datasetName': ,
    'dataResourceName': 
});
```

Implement for both:
- [x] ckan 2.8 (`main` branch)
- [x] ckan 2.9 (`ckan_2.9_upgrade` branch)

PR #244, opened 2022-10-19
GA download tracker (CKAN 2.8)

KatiRG:
## What this PR accomplishes
Implements a new JS function `trackDownload()` that is triggered every time a download button or download URL is clicked. The dataLayer object is sent to Google analytics for further processing.

## Issue addressed
- #243
- small fix in passing group  name for #237 in `ckanext/ontario_theme/templates/internal/base.html` and `ckanext/ontario_theme/templates/internal/gtm_pageview.html`)

## What needs review
The functionality of the code has already been reviewed from the Google analytics side. The reviewer here can generally review the code. The CKAN docs suggest creating a [JS module](https://docs.ckan.org/en/2.9/contributing/frontend/javascript-module-tutorial.html#building-a-javascript-module), but `ckan.module` was not recognized and as an alternative, an ordinary JS function was created instead.


KatiRG (commented):
Approving based on approval from external reviewer

PR #245, opened 2022-10-19
GA download tracker (CKAN 2.9)

KatiRG:
## What this PR accomplishes
Implements a new JS function `trackDownload()` that is triggered every time a download button or download URL is clicked. The dataLayer object is sent to Google analytics for further processing.

This PR is the CKAN 2.9 version of PR #244 (CKAN 2.8). The main difference is how assets are defined. In CKAN 2.9, assets are defined with `{% asset %}` tags in the view templates, and they need to be included in the `ckanext/ontario_theme/fanstatic/internal/webassets.yml` file.

## Issue addressed
- #243
- small fix in passing group  name for #237 in `ckanext/ontario_theme/templates/internal/base.html` and `ckanext/ontario_theme/templates/internal/gtm_pageview.html`)

## What needs review
As in PR #244, the reviewer here can generally review the code.  The CKAN docs suggest creating a [JS module](https://docs.ckan.org/en/2.9/contributing/frontend/javascript-module-tutorial.html#building-a-javascript-module), but `ckan.module` was not recognized and as an alternative, an ordinary JS function was created instead.

### DIFF
--- a/ckanext/ontario_theme/fanstatic/internal/ontario_theme_download_tracker.js
+++ b/ckanext/ontario_theme/fanstatic/internal/ontario_theme_download_tracker.js
@@ -7,15 +7,6 @@ function trackDownload(resourceUrl, orgName, pkgTitle, groupName) {
 
   let fileName = resourceUrl.split('download/')[1];
 
-  if (groupName.length > 2) { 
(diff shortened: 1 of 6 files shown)


abi70500 (commented):
Everything looks good ,you can go head

PR #248, opened 2022-11-29
Update breadcrumb styling to match Ontario Design System (CKAN 2.8)

KatiRG:
## What this PR accomplishes
This PR is for CKAN 2.8. A similar PR will follow for CKAN 2.9.

- adds breadcrumb styling from the Ontario Design System into a common css file in the `fanstatic` root directory and removes breadcrumb styling from internal and external `ontario_theme.css` files 
- defines the `fanstatic` root directory in `plugin.py` which is needed for importing the css file
- imports the common css file in the `base.html` page of the internal site only
- overrides the `page.html` file from the CKAN codebase and uses the `page.html` file in this theme, where the components use the classes in the ontario theme css file. Also, the home icon is removed.

## Issues addressed
- breadcrumb styling matches Design System (but application font is still Open Sans)
- breadcrumb styling moved to a common css file in the fanstatic root directory that is accessible to both internal and external sites
-  http link to google fonts replaced by local file in fanstatic root directory

## What needs review
- please check both internal and external sites for the new breadcrumb styles
- look over the common css file for possible redundancies, inefficiencies, etc.


ckanext/ontario_theme/fanstatic/fonts_googleapis.css
```diff
@@ -0,0 +1,721 @@
+/* cyrillic-ext */
```
aimalkhan, line 1: Do we need to include the google fonts, instead of having the client pull them from the url (delete line of code in `ckanext/ontario_theme/templates/internal/base.html`)

aimalkhan (approved):
a couple of comments needing discussion, otherwise looks good!

PR #249, opened 2022-11-29
Update breadcrumb styling to match Ontario Design System (CKAN 2.9)

KatiRG:
## What this PR accomplishes
This PR accomplishes the same as PR #248, but for CKAN 2.9. The code changes are not the same because of the way assets are defined: CKAN 2.9 uses `.yml` files.

Differences:
- a `ckanext/ontario_theme/fanstatic/webassets.yml` file created to define location of the css files in the `fanstatic` root dir
- the common css files are also defined in `ckanext/ontario_theme/fanstatic/internal/webassets.yml`

Commonalities:
- common css file for breadcrumbs stored in `fanstatic` root directory and removed from internal and external `ontario_theme.css`
- `fanstatic` root directory defined in `ckanext/ontario_theme/plugin.py` for both internal and external sites
- the css assets need only be imported in `ckanext/ontario_theme/templates/internal/base.html` and not also in the external `base.html`

## Issues addressed
- breadcrumb styling matches Design System (but application font is still Open Sans)
- breadcrumb styling moved to a common css file in the `fanstatic` root directory that is accessible to both internal and external sites
- http link to google fonts replaced by local file in `fanstatic` root directory

## What needs review
- confirm that the breadcrumb style changes are in place
- confirm that the application font does not change now that it is loaded from a local file
Note: only the external site can be tested for now

### DIFF
--- a/ckanext/ontario_theme/templates/internal/snippets/home_breadcrumb_item.html
+++ b/ckanext/ontario_theme/templates/internal/snippets/home_breadcrumb_item.html
@@ -1,4 +1,3 @@
-{# Used to insert the home icon into a breadcrumb #}
 <li class="home">
     <a href="{{ h.url_for('home.index') }}" aria-label="{{ _('Home') }}">
         {{ _('Home') }}
--- a/ckanext/ontario_theme/templates/internal/snippets/home_breadcrumb_item.html
+++ b/ckanext/ontario_theme/templates/internal/snippets/home_breadcrumb_item.html
@@ -1,0 +1,6 @@
+{# Used to insert the home icon into a breadcrumb #}
+<li class="home">
+    <a href="{{ h.url_for('home.index') }}" aria-label="{{ _('Home') }}">
+        {{ _('Home') }}
+    </a>
+</li>
--- a/ckanext/ontario_theme/fanstatic/common.css
+++ b/ckanext/ontario_theme/fanstatic/common.css
@@ -1,0 +1,32 @@
+/* ========================================================================
+   Breadcrumbs - à la Design System
+   ======================================================================== */
+.toolbar ol.breadcrumb {
+  color: #aaa;
+  padding: 10px 16px;
+  list-style: none;
+}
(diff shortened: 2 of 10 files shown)


Alexandra-Green (commented):
On the About and Help page the breadcrumb is aligned with the content on the page. On the Dataset, Organization, and Group pages, the breadcrumb is right of the left margin of the page, is this something that can be fixed?

ckanext/ontario_theme/templates/internal/snippets/home_breadcrumb_item.html
```diff
@@ -0,0 +1,6 @@
+{# Used to insert the home icon into a breadcrumb #}
```
Alexandra-Green, line 1: Will we be inserting the home icon?

Alexandra-Green:
This may not be within the scope of this ticket, but why do the last breadcrumbs link to the current page?

Alexandra-Green:
✅ breadcrumb style changes are in place
✅ the application font does not change now that it is loaded from a local file

Alexandra-Green (approved):

KatiRG:
> On the About and Help page the breadcrumb is aligned with the content on the page. On the Dataset, Organization, and Group pages, the breadcrumb is right of the left margin of the page, is this something that can be fixed?

This is not a breadcrumb issue: the left margin of the breadcrumbs remains the same on all pages. It's the main content that starts at a different margin on the About and Help pages, I am guessing because they don't have a searchbar. This would be up to the designer to decide if this should be changed.

KatiRG:
> This may not be within the scope of this ticket, but why do the last breadcrumbs link to the current page?

That's another question for the designer. This is how it is currently.

ckanext/ontario_theme/templates/internal/snippets/home_breadcrumb_item.html
KatiRG: No, the home icon was removed. That was a legacy comment, I'll delete it, thanks!

PR #250, opened 2022-12-02
added Ontario Forest Region to Geographic Breakdown choices

rleblanc-on:
MNRF requested 'Ontario Forest Region' be added to the list of possible 'Geography Breakdowns'.

### DIFF
--- a/ckanext/ontario_theme/schemas/external/ontario_theme_dataset.json
+++ b/ckanext/ontario_theme/schemas/external/ontario_theme_dataset.json
@@ -205,7 +205,14 @@
             "en": "Forward Sortation Area",
             "fr": "régions de tri d’acheminement"
           }
-        }
+        },
+        {
+          "value": "ofa",
+          "label": {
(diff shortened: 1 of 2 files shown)


KatiRG:
@rleblanc-on works great! Did you also want it to be an option for the internal site? If so, then I think you also need to add the changes to the json file in `ckanext-ontario_theme/ckanext/ontario_theme/schemas/internal`.

aimalkhan (approved):
👍

aimalkhan:
> @rleblanc-on works great! Did you also want it to be an option for the internal site? If so, then I think you also need to add the changes to the json file in `ckanext-ontario_theme/ckanext/ontario_theme/schemas/internal`.
> 
> Nevermind @rleblanc-on ! We can handle the internal site later, as we have no way of testing ckan 2.9 internally yet.

I would suggest we do the change any way, otherwise we might miss it in the future.

KatiRG (approved):

PR #257, opened 2022-12-12
Ckan 2.9 upgrade topics section

Alexandra-Green:
## What this PR accomplishes

- Adds method to return the count of each keyword's datasets and added to homepage in `plugin.py`
- Reduces size of topic icons

## Issue(s) addressed

- Updates the “Category” heading to be “Topics”
- Changes the size of the topic icons to be smaller
- The number of datasets for each topics is displayed after the topic name

## What needs review

- If the method is redundant. `ontario_theme/templates/internal/snippets/facet_list.html` generates the datasets of each keyword differently, but I couldn't understand how to use it this way.
- If the dataset counts are accurate
- If section matches Figma


KatiRG:
The "What's new" section should be removed, according to the figma. But one of your previous PRs removes this section, so when you get to re-submitting this PR, just make sure you don't add it back in.

ckanext/ontario_theme/plugin.py
```diff
@@ -369,14 +369,13 @@ def get_group(group_id):
...
+def get_keyword_count(keyword):
+    '''Helper to return the dataset count of the indicated group
     '''
-    recently_updated_datasets = toolkit.get_action('package_search')(
-        data_dict={'rows': 3,
-                    'sort': 'current_as_of desc'})
-    return recently_updated_datasets['results']
-
+    tags = 'tags:"{}"'.format(keyword)
+    keyword_count = toolkit.get_action('package_search')(
```
KatiRG, line 376: I see an issue here: by counting only the number of english keywords, this could cause a mis-match with the counts on the Dataset page when in French, in the case where an English keyword was entered but the French keyword was left blank. To get around this, try to find how keywords are counted in the dataset page and return that in the `get_keyword_count()`. Below are screenshots of the Dataset page in English and French showing 2 counts for Health in English and 1 count in French. On the homepage, 2 counts will be shown in both English and French, but then clicking on the Health icon only shows 1 dataset when navigating in French.

Screen Shot 2022-12-15 at 10 43 26 AM
Screen Shot 2022-12-15 at 10 44 01 AM

Alexandra-Green: > ontario_theme/templates/internal/snippets/facet_list.html generates the datasets of each keyword differently, but I couldn't understand how to use it this way.
It's in this location if you want to take a look at it.

KatiRG: @Alexandra-Green I couldn't find how to define `items` in that snippet in our html file either. Here's something else that works that you could try:

- in [ontario_theme_categories.html](https://github.com/ongov/ckanext-ontario_theme/blob/ckan_2.9_upgrade_topics_section/ckanext/ontario_theme/templates/internal/home/snippets/ontario_theme_categories.html#L44), set the language before the `for` loop: `{% set language = h.lang() %}`
- in [ontario_theme_categories.html L53](https://github.com/ongov/ckanext-ontario_theme/blob/ckan_2.9_upgrade_topics_section/ckanext/ontario_theme/templates/internal/home/snippets/ontario_theme_categories.html#L53), instead of passing in the english keyword to your function, pass the keyword in the language that the application is currently in, and also pass the language:

```
({{ h.ontario_theme_get_keyword_count(_(topics[i]['label']), language) }})
```

- modify [get_keyword_count](https://github.com/ongov/ckanext-ontario_theme/blob/ckan_2.9_upgrade_topics_section/ckanext/ontario_theme/plugin.py#L372) to count the keywords as per the language passed: 
```
def get_keyword_count(keyword_lang, language):
    '''Helper to return the dataset count of the indicated group by language
    '''
    facet = "keywords_{}".format(language)
    package_keywords = toolkit.get_action('package_search')(
    data_dict={'facet.field': [facet],
        'rows': 0})
    active_keywords = package_keywords['facets'][facet]
    if keyword_lang in active_keywords:
        keyword_count_by_lang = active_keywords[keyword_lang]
    else:
        keyword_count_by_lang = 0

    return keyword_count_by_lang
```

ckanext/ontario_theme/templates/internal/home/snippets/ontario_theme_categories.html
```diff
@@ -50,6 +51,7 @@
     <div class="col-md-2-10">
       <i class="fa fa-{{topics[i]['fa-icon']}}"></i> <br>
         {{ _(topics[i]['label']) }}
+        ({{ h.ontario_theme_get_keyword_count(topics[i]['label'], language) }})
```
KatiRG, line 54: You have to pass `_(topics[i]['label'])` with the underscore

KatiRG (approved):

PR #264, opened 2022-12-23
CKAN 2.9 search bar with DS

KatiRG:
## What this PR accomplishes
- Updates the search bar component called by dataset page, organization page, and groups page by to use the DS component with input arguments
- Modifies DS search bar for home page to use input arguments

Note:
- styling is out of the scope of this PR
- there is a known bug in the search bar focus of the Groups page, so can be ignored

## Issue addressed
- ticket to replace search bar box with DS and adapt for re-usability

## What needs review
- Check that search bar works in dataset page, organization page, and groups page and is consistent with DS
- Check that search bar still works in home page


Alexandra-Green:
Could the homepage search bar use the `ckanext/ontario_theme/templates/internal/snippets/search_form.html`? 

I see the main difference is that the homepage search bar uses `action` in `<form>`

Alexandra-Green (commented):
✅ Check that search bar works in dataset page, organization page, and groups page and is consistent with DS
✅ Check that search bar still works in home page

KatiRG:
> Could the homepage search bar use the `ckanext/ontario_theme/templates/internal/snippets/search_form.html`?
> 
> I see the main difference is that the homepage search bar uses `action` in `<form>`

This would be a good idea but could not get it to bypass the other elements on the page. We could make this a separate issue to be worked on when there is more time.

Alexandra-Green (approved):

PR #273, opened 2023-01-10
FIX: zip files containing directories can be uploaded

KatiRG:
## What this PR accomplishes
Mimetypes of every item inside a zip file to be uploaded were being checked, but the case where the zip file contains a directory was not accounted for. To get around this, a condition was added to ignore strings that represent directories, and only the mimetypes of filename strings are now checked.

## Issue resolved
Fixes error in trying to upload zip files containing directories.

## What needs review
- Please confirm that zip files containing directories can be uploaded successfully.
- If there is a better way of determining whether an item is a directory name, please suggest it. I am simply checking whether there is a `/` at the end of the string.


aimalkhan (approved):
looks good 👍

PR #274, opened 2023-01-11
CKAN2.8 FIX: allows uploads of zip files containing folders

KatiRG:
## What this PR accomplishes
Same as #273 but for the CKAN 2.8 branch.
Code change is same.


aimalkhan (approved):
looks good 👍

PR #279, opened 2023-01-18
Ckan 2.9 Design System header changes

Alexandra-Green:
## What this PR accomplishes

- Removing “Groups” 
- Implementing the [Design System's application header](https://designsystem.ontario.ca/components/detail/application-header.html) with a search bar on all screen sizes 
- Updating “Organizations” to be “Ministries”
- Underline the links to indicate they are clickable

## Issue(s) addressed

- black bar with:
   - smaller version of the Ontario logo
   - language toggle
- secondary grey application bar with: 
   - application name
   - navigation links
   - site menu
   - search bar
- Adds directory for javascript files
 
## What needs review

- Please advise if a DS folder should be created to put files 12col-grid.css and font_ds.css in
- ### Search bar: 
   - Responsive
   - Works (redirects to datasets page with user input as query)
   - The reset (x) appears after user inputs a key
   - clicking the reset (x) clears the field
   - the reset (x) does not overlap with long user inputs
   - On medium and mobile, clicking the search icon, opens the search bar
   - only the search bar, application name, and close svg should be visible on medium and only the search bar and close svg are visible on mobile
   - clicking the close svg closes the search bar
   - All focus/active/hover states match the figma
- ### Navigation tabs: 
   - All focus/active/hover states and cursors match the figma
   - the current page is underlined
   - clicking the tabs redirects to the correct page
   - Groups is removed
- ### Menu:
   - All focus/active/hover states and cursors match the figma
   - the current page css matches figma
   - clicking the menu button opens the dropdown and shows an (x)
   - clicking the links redirects to the correct page
   - tablet and mobile designs match figma
- ### Translation:
   - toggling the language works and all header text is translated appropriately
   - using the navigation bar and dropdown menu redirects to the correct page
   - toggle button focus/hover/active states matches figma
   - clicking ontario in the top left will navigate to the correct page depending on the site's current language
- Clicking the application name goes to the homepage

#### NOTES:

- Adding 3 nav tabs in medium with the search icon created a lot of problems with overlapping and overflow
- "Menu" text is hidden from view before the mobile screen size due to overlapping and overflow issues
- The header uses the Design System's grid system, while the rest of the site uses CKAN's, consequently the header is slightly narrower than the homepage.


KatiRG (changes requested):
In Firefox under mobile mode, the
Screen Shot 2023-01-30 at 9 37 03 AM
 hamburger menu is cut off (but it's ok in Chrome):

KatiRG (changes requested):
Focus highlights correctly in blue but the button should become a lighter gray as per the figma and the design system
Screen Shot 2023-01-30 at 9 54 10 AM
Screen Shot 2023-01-30 at 9 54 45 AM

KatiRG (changes requested):
If in smaller screen size with menu open, then expanding screen size, there is a point where the open menu should disappear like the closed menu does (replaced by the horizontal menu items). But when the menu is open it remains on screen along with the horizontal menu items.

Open menu, smaller screen:
Screen Shot 2023-01-30 at 10 36 57 AM

larger width, horizontal menu items appear but the open dropdown menu does not disappear and it overlaps with the search bar:

Screen Shot 2023-01-30 at 10 38 12 AM

Alexandra-Green:
> If in smaller screen size with menu open, then expanding screen size, there is a point where the open menu should disappear like the closed menu does (replaced by the horizontal menu items). But when the menu is open it remains on screen along with the horizontal menu items.

This has been changed

Alexandra-Green:
> The focus (with tab button) on the menu item should be a darker gray and also underlined according to the figma and DS page

Has been changed

Alexandra-Green:
> Focus highlights correctly in blue but the button should become a lighter gray as per the figma and the design system 

This has been changed

Alexandra-Green:
> In Firefox under mobile mode, the hamburger menu is cut off (but it's ok in Chrome)

This has been changed, but now there are issues with the menu button. I will have to look at this further.

ckanext/ontario_theme/fanstatic/internal/webassets.yml
```diff
@@ -8,6 +8,7 @@ ontario_theme_common:
     - common.css
     - fonts_googleapis.css
```
KatiRG, line 9: @Alexandra-Green Do we still need `fonts_googleapis.css` now that we have all the font files in `public/fonts/ds-fonts/`?

KatiRG (changes requested):
The Settings icon when logged in still says "Settings" when toggled to French, and in the hover text as well.
Screen Shot 2023-02-03 at 12 03 27 PM

KatiRG:
The order of tab focus on the admin bar seems to be in reverse: it starts with Logout, then proceeds left through Settings, Dashboard, etc. I think it makes more sense for the first focus to be on the user icon, then proceed rightwards.

Screen Shot 2023-02-03 at 1 30 26 PM

ckanext/ontario_theme/fanstatic/common.css
```diff
@@ -253,6 +247,14 @@ input[type=reset].ontario-search-reset:focus{
   .module-content {
     padding: 0px; /* brings top of file list closer to Order By menu */
   }
+  
+  .search-form {
+    padding: 20px;
+    background: #fafafa;
+  }
```
KatiRG, line 254: `.search-form` can be removed because its properties are over-ruled:

Screen Shot 2023-02-06 at 9 26 28 AM

ckanext/ontario_theme/fanstatic/common.css
```diff
@@ -435,6 +437,516 @@ a:focus {
...
+}
+@media screen and (max-width: 73em) {
+  .ontario-navigation--open .ontario-application-subheader-menu__container {
+    position: relative;
+    z-index: 3;
+  }
+  .ontario-navigation a {
+    padding: calc(0.75rem - (0.25rem / 3.25)) 2rem;
+  }
+  .account-text{
```
KatiRG, line 790: `account-text` does not seem to be called in the application for the admin menu items. Actually removing both classes `class="text account-text"` from the span in `header.html` (`{% block header_account_container_content %}`) doesn't change the appearance of the texts.

ckanext/ontario_theme/fanstatic/internal/webassets.yml
Alexandra-Green: It doesn't look like we need it anymore, I'll remove it

ckanext/ontario_theme/fanstatic/common.css
KatiRG: @Alexandra-Green there is already a `.search-form` class on line 200:
```
.search-form {
    display: flex;
    flex-direction: column;
  }
```
Can you consolidate them? Also, please check if `padding: 20px` is ever used; in individual group and ministry pages, it is crossed out:

Screen Shot 2023-02-08 at 8 29 44 AM

KatiRG (changes requested):

ckanext/ontario_theme/fanstatic/common.css
Alexandra-Green: I consolidated it, the padding is necessary for the top and sides. The id of the form has styling properties to remove the bottom padding.

ckanext/ontario_theme/fanstatic/external/ontario_theme.css
```diff
@@ -1083,32 +1087,39 @@ textarea[readonly]::placeholder {
...
   font-size: 14px;
+  text-decoration: none;
+  padding: 0.5rem 0.75rem;
 }
 /* The user account info text and icons, when the user's pointer is hovering
    over them. */
-.account-masthead .account ul li a:hover {
-  color: #eee;
-  background-color: #333;
+.account ul li a:hover {
```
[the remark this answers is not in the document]
Alexandra-Green: I'm not sure I understand your question, but I can change the colour if it blends too well into the background in the hover state and non hover states.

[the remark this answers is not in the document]
KatiRG: Yes that's what I meant. The designers haven't given us admin panel designs yet, so we can make it the same colour as the background and submit it for review.

PR #280, opened 2023-01-20
Ckan 2.9 French breadcrumb translation

Alexandra-Green:
## What this PR accomplishes

- Translates the breadcrumb for Ministries when in dataset pages

## What needs review

- That the breadcrumb for Ministries is translated when toggled in French mode, specifically when navigating to an individual dataset from a Ministry or from the dataset search page


KatiRG (changes requested):
The problem persists in the test environment.
What about removing the call to `pkg.organization` title and taking the organization title only from `organization_dict` ?

e.g. `{% set organization = h.get_translated(organization_dict, 'title') or organization_dict['title'] %}` 

All ministries are (should be?) forced to supply their French name equivalents, but I have not tested if this suggested change works on all pages.

KatiRG (approved):
Looks ok, I couldn't find any places where this fails.

PR #282, opened 2023-02-02
allow only write-access users to see resources in non-open pkgs

KatiRG:
## What this PR accomplishes
If a package is not open (i.e. restricted or under review), users who are do not have write access (i.e. users who are not sysadmins or org admins) should not be able to see the resource list. 

## What needs review
Please confirm that the resource list of a package with access level that is not open is only displayed for sysadmins or org admins.


aimalkhan (approved):
👍 looks good

PR #286, opened 2023-02-17
Ckan 2.9 Improved design for keyword texts

Alexandra-Green:
## What this PR accomplishes

Implements new pattern for displaying the keywords & selected filters:

## Issue(s) addressed

- French keyword filter translations
- Moves filters to below search query text
- Selected words in search query text are now bold
- Adds the close (x) and functionality to each search bar (including home) across site
- Adjusts Dataset and Ministry pages search labels, translations, and placeholders according to Figma design
- Shortens long input text placeholders in mobile and tablet across the site
- Allows empty searches

## What needs review

- Each search bar should clear the input field on clicking the (x), regardless of if there is a search query or not
- long placeholders are shortened in mobile and tablet
- The correct words are bold in the search results text and all translations are correct
- filter pills are correctly translated and look fine in mobile
- search bars align with Figma design

Notes: 

- clicking the "datasets with data" filter does not currently show a pill. This filter is going to be removed with the dataset search page changes, but let me know if there should be a pill.
- I added the `odc-search.js` script, let me know if I should consolidate this into the `ontario-header.js` or leave as is. 
- `facet_list.html` and `search_form.html` now use some of the same code to translate the filter pills, should I be putting this in a separate html file and calling it as a snippet?

PR #287, opened 2023-02-21
Fix broken sort and define different default sort order

KatiRG:
## What this PR accomplishes
1. Defines a helper function `sort_page_items()` to alphabetically sort dataset names and ministry names in English or French _before_ they are looped through on the display pages (`organization_list.html` and `package_list.html`).
2. Overrides CKAN's default sort order `'score desc, metadata_modified desc'` to `'metadata_modified desc'` (i.e. **Last modified**) in the Ontario theme templates for datasets. The default sort order is invoked when `request.params.get('sort')` is empty, i.e., when first landing on the page. A sort order subsequently selected from the **Order by** drop-down menu after landing on the page will be stored in `request.params.get('sort')` and will then be used to sort. (For reference: CKAN's default sort order is defined in the core CKAN [package_search() function in get.py](https://github.com/ckan/ckan/blob/ckan-2.9.7/ckan/logic/action/get.py#L1842).) 

## Issues addressed
1. broken sort: **Name ascending** and **Name descending** do not work properly for Ministries and Datasets, in both languages
3. change default sort order to:
- **Name ascending** as default for Organizations (Ministries) and Groups
- **Last modified** as default for Datasets

## What needs review
- verify that Ministries are correctly ordered by **Name ascending** and **Name descending** in English and French
- verify that datasets are correctly ordered by **Last modified** / **Modifié le** when first landing on either the Datasets page or when clicking on a Ministry in the Ministries page to open a list of datasets for that ministry

Note: In French, accented first characters such as `É` are alphabetically after `A-Z` characters.

### DIFF
--- a/ckanext/ontario_theme/plugin.py
+++ b/ckanext/ontario_theme/plugin.py
@@ -394,6 +394,18 @@ def get_keyword_count(keyword_lang, language):
 
     return keyword_count_by_lang
 
+def sort_page_items(obj, fields_dict):
+    '''Helper to return the page items object sorted by title_translated
+    for the current language.
+    '''
+    field = fields_dict['field']
+    lang = fields_dict['lang']
+    reverse = fields_dict['reverse']
+    return sorted(obj, 
+                  key=lambda x: x[field][lang], 
+                  reverse=reverse)
+
+
 def get_popular_datasets():
     '''Helper to return most popular datasets, based on ckan core tracking feature
     '''
@@ -720,7 +732,8 @@ type data_last_updated
                 'ontario_theme_home_block_image': home_block_image,
                 'ontario_theme_home_block_link': home_block_link,
                 'ontario_theme_get_group_datasets': get_group_datasets,
-                'ontario_theme_get_keyword_count': get_keyword_count
+                'ontario_theme_get_keyword_count': get_keyword_count,
+                'ontario_theme_sort_page_items': sort_page_items
                 }
 
     # IBlueprint
--- a/ckanext/ontario_theme/templates/internal/organization/read.html
+++ b/ckanext/ontario_theme/templates/internal/organization/read.html
@@ -21,6 +21,9 @@ Override CKAN's read.html to add some new facets.
     (_('Last Modified'), 'metadata_modified desc'),
     (_('Popular'), 'views_recent desc') if g.tracking_enabled else (false, false) ]
   %}
+  {% if not sort_by_selected %}
+    {% set sort_by_selected = 'metadata_modified desc' %}
+  {% endif %}
   {% snippet 'snippets/search_form.html', form_id='organization-datasets-search-form', type='dataset', query=q, sorting=sorting, sorting_selected=sort_by_selected, count=page.item_count, facets=facets, placeholder=_('Search datasets'), show_empty=request.params, fields=fields %}
 {% endblock %}
 
--- a/ckanext/ontario_theme/templates/internal/package/search.html
+++ b/ckanext/ontario_theme/templates/internal/package/search.html
@@ -62,6 +62,7 @@ in the core template for search.html.
     <div class="module-content">
       {% block package_search_results_list %}
         {{ super() }}
+        {{ h.snippet('snippets/package_list.html', packages=packages) }}
       {% endblock %}
     </div>
 
--- a/ckanext/ontario_theme/templates/internal/package/search.html
+++ b/ckanext/ontario_theme/templates/internal/package/search.html
@@ -48,7 +48,13 @@ in the core template for search.html.
       {% set type = dataset_type %}
       {% set searchbar_title = _('Search ' + dataset_type + 's') + _(':') %}
       {% set hidden_searchbar_title = _('Search ' + dataset_type + 's') %}
-      {% snippet 'snippets/search_form.html', searchbar_title=searchbar_title, hidden_searchbar_title=hidden_searchbar_title, form_id=form_id, type=type, query=query, sorting=sorting, sorting_selected=c.sort_by_selected, count=c.page.item_count, placeholder=placeholder, sr_input_text=sr_input_text, sr_submit_text=sr_submit_text, facets=facets, show_empty=request.params, error=c.query_error, fields=c.fields %}
+      <!-- override default sort order 'score desc, metadata_modified desc' to 'metadata_modified desc' -->
+      {% set default_order = 'metadata_modified desc' %}
+      <!-- define sort_order to default_order when first landing on page, otherwise
+           set to user menu selection stored in request.params['sort']
+       -->
+      {% set sort_order = default_order if not request.params else request.params['sort'] %}
+      {% snippet 'snippets/search_form.html', searchbar_title=searchbar_title, hidden_searchbar_title=hidden_searchbar_title, form_id=form_id, type=type, query=query, sorting=sorting, sorting_selected=sort_order, count=c.page.item_count, placeholder=placeholder, sr_input_text=sr_input_text, sr_submit_text=sr_submit_text, facets=facets, show_empty=request.params, error=c.query_error, fields=c.fields %}
     {% endblock %}
   </div>
 {% endblock %}
--- a/ckanext/ontario_theme/templates/internal/organization/index.html
+++ b/ckanext/ontario_theme/templates/internal/organization/index.html
@@ -50,7 +50,12 @@
   {% block organizations_list %}
     {% if c.page.items or request.params %}
       {% if c.page.items %}
-        {% snippet "organization/snippets/organization_list.html", organizations=c.page.items %}
+        {# Custom sort by language for name ascending and descending #}
+        {% set current_lang = request.environ.CKAN_LANG %}
+        {% set reverse = True if 'desc' in request.params['sort'] else False %}
+        {% set fields_dict = {'field': 'title_translated','lang': current_lang, 'reverse': reverse} %}
+        {% set sorted_org = h.ontario_theme_sort_page_items(c.page.items, fields_dict) %}
+        {% snippet "organization/snippets/organization_list.html", organizations=sorted_org %}
       {% endif %}
     {% else %}
       <p class="empty">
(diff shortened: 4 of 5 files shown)


ckanext/ontario_theme/plugin.py
```diff
@@ -394,6 +394,18 @@ def get_keyword_count(keyword_lang, language):
...
 
+def sort_page_items(obj, fields_dict):
+    '''Helper to return the page items object sorted by title_translated
+    for the current language.
+    '''
+    field = fields_dict['field']
+    lang = fields_dict['lang']
+    reverse = fields_dict['reverse']
+    return sorted(obj, 
+                  key=lambda x: x[field][lang],
```
Alexandra-Green, line 405: I'm getting a `KeyError: 'en'` from this line, when trying to go to the next page and when I try choosing the order by "Name Ascending" option on the Ministries search page.

ckanext/ontario_theme/templates/internal/organization/read.html
```diff
@@ -21,6 +21,9 @@
     (_('Last Modified'), 'metadata_modified desc'),
     (_('Popular'), 'views_recent desc') if g.tracking_enabled else (false, false) ]
   %}
+  {% if not sort_by_selected %}
+    {% set sort_by_selected = 'metadata_modified desc' %}
```
Alexandra-Green, line 25: It's being sorted by last modified but the order by drop down is still defaulting to display "Relevance"

Alexandra-Green (changes requested):
Since the sort is being done according to ASCII codes, uppercase letters < lowercase letters. Datasets like `COVID-19 Vaccine Data in Ontario` appear before `Confirmed positive cases of COVID-19 in Ontario` when sorted by "Name Ascending" even though it should come after (n should go before v, alphabetically)

KatiRG:
> Since the sort is being done according to ASCII codes, uppercase letters < lowercase letters. Datasets like `COVID-19 Vaccine Data in Ontario` appear before `Confirmed positive cases of COVID-19 in Ontario` when sorted by "Name Ascending" even though it should come after (n should go before v, alphabetically)

There are different rules we can choose, there is no absolute correct way. Does ontario.ca use one convention vs the other? Do we have any style guide to reference?

ckanext/ontario_theme/plugin.py
KatiRG: I added a check for empty `title_translated` fields and no longer see this error. Please check.

ckanext/ontario_theme/templates/internal/snippets/facet_list.html
```diff
@@ -31,7 +31,7 @@ <h2 class="module-heading">
     <nav>
       <ul class="{{ nav_class or 'list-unstyled nav nav-simple nav-facet' }}">
         {% for item in items %}
-          {{ h.snippet('snippets/translate_facets.html', fields=fields, name=name, item=item, scheming_choices=scheming_choices, extras=extras) }}
+          {{ h.snippet('snippets/translate_facets.html', fields=fields, name=name, item=item, scheming_choices=scheming_choices) }}
```
KatiRG, line 34: Why is `extras=extras` removed?

ckanext/ontario_theme/templates/internal/snippets/search_form.html
```diff
@@ -69,13 +69,13 @@ <h2 aria-label='{{ hidden_searchbar_title }}'><p class="theme_p" aria-hidden="tr
...
 
-            {{ h.snippet('snippets/translate_facets.html', fields=schema_dict, name=item_id, item=item, scheming_choices=choices, extras=extras) }}
+            {{ h.snippet('snippets/translate_facets.html', fields=schema_dict, name=item_id, item=item, scheming_choices=choices) }}
 
               {% if item.active %}
                 <span class="filtered pill">
                   <span class="pill-contents">
                     {{item.display_name}}
-                    <a href="{{ h.remove_url_param(item_id, item.name, extras=extras, alternative_url=alternative_url) }}" class="remove" title="{{ _('Remove') }}"><i class="fa fa-times"></i></a>
+                    <a href="{{ facets.remove_field(field, item.name) }}" class="remove" title="{{ _('Remove') }}"><i class="fa fa-times"></i></a>
```
KatiRG, line 78: also here why are `extras` removed?

Alexandra-Green: This was the line that was causing the error. I removed the h.remove_url_param and changed it for facets.remove_field. This was the original code, and that's why it is working on stage and not on test.

ckanext/ontario_theme/templates/internal/snippets/facet_list.html
Alexandra-Green: when I created `{{ h.snippet('snippets/translate_facets.html', fields=fields, name=name, item=item, scheming_choices=scheming_choices, extras=extras) }}` in #286 I added in extras when I really didn't have to. I wasn't aware of it's purpose then, but I understand it now, and it's not needed in the line.

Alexandra-Green (changes requested):
I'm getting and Error [500] when trying to create a new ministry

ckanext/ontario_theme/plugin.py
```diff
@@ -394,6 +394,187 @@ def get_keyword_count(keyword_lang, language):
...
+    for (param, value) in OrderedDict(request_params_items).items():
+        if param not in ['q', 'page', 'sort'] \
+            and len(value) and not param.startswith('_'):
+            if not param.startswith('ext_'):
+                fq += ' %s:"%s"' % (param, value)
+            else:
+                search_extras[param] = value
+    
+    if current_org:
+        fq += ' %s:"%s"' % ('organization', current_org)
```
Alexandra-Green, line 486: Sort resolution isn't currently applied to groups. Sorting by name ascending/descending on an individual groups page will sort the correct number of datasets but from the entire list of datasets, not the list of datasets belonging to that group.

ckanext/ontario_theme/templates/internal/organization/read.html
```diff
@@ -21,13 +21,37 @@
...
+                sorting_selected=sort_by_selected, 
+                count=page.item_count, 
+                facets=facets, 
+                placeholder=searchbar_text, 
+                show_empty=request.params, 
+                fields=fields %}
   </form>
 {% endblock %}
+{% block packages_list %}
+    {% if page.items %}
```
Alexandra-Green, line 50: I think the Error [500] is due to no else in this if statement when there is no packages.

ckanext/ontario_theme/templates/internal/package/search.html
```diff
@@ -46,10 +46,37 @@ <h1 class="ontario-h1">{% block page_heading %}{{ _(dataset_type + 's') |capital
...
+      #}
+      {% if request['params'] %}
+        {% if 'sort' in request['params'] %}
+          {% set sort_order = request.params['sort'] %}
+        {% else %}
+          {% set sort_order = 'metadata_modified desc' %}
+        {% endif %}
+      {% else %}
+        {% set sort_order = 'metadata_modified desc' %}
+      {% endif %}
```
Alexandra-Green, line 60: Can this be combined into 1 if statement?

KatiRG:
> I'm getting and Error [500] when trying to create a new ministry & if I try clicking a new ministry that doesn't have any datasets, I am also getting the Error [500]

Should be resolved now

ckanext/ontario_theme/plugin.py
KatiRG: Should be resolved now

Alexandra-Green (changes requested):
I'm getting an Error [500], when going to the groups tab on a dataset page, neither as a regular user or an admin.

The groups page says relevance on loading, instead of last modified, but that can be moved to a new ticket for after launch.

KatiRG:
> I'm getting an Error [500], when going to the groups tab on a dataset page, neither as a regular user or an admin.
> 
> The groups page says relevance on loading, instead of last modified, but that can be moved to a new ticket for after launch.

Good catch! Both are addressed now and should close #287

ckanext/ontario_theme/plugin.py
```diff
@@ -394,6 +394,202 @@ def get_keyword_count(keyword_lang, language):
...
+    # to 'item_count'.
+    package_search=toolkit.get_action('package_search')(
+                data_dict={
+                        'q': q,
+                        'fq': fq.strip(),
+                        'start': 0,
+                        'sort': 'title desc',
+                        'rows': item_count,
+                        'extras': search_extras,
+                        'include_private': True
```
Alexandra-Green, line 517: > private datasets don't show up in any search

Doesn't this line in the code negate this ^

Alexandra-Green (approved):
✅ Ministries are correctly ordered by Name ascending and Name descending in English and French
✅ datasets are correctly ordered by Last modified / Modifié le when
        first landing on either the Datasets page
        ❓ when clicking on any facet (such as a keyword) to open a list of datasets
        when clicking on a Ministry in the Ministries page to open a list of datasets for that ministry
        when clicking on an individual Group
✅ Groups are correctly ordered by Name ascending and Name descending in English and French
✅ searching inside a particular Group does not return results from datasets that do not belong to the Group
❓ private datasets don't show up in any search (see comment) 

> when clicking on any facet (such as a keyword) to open a list of datasets

Does this mean that when a user is sorting by name ascending and clicks a filter, it should then be sorted by last modified? 


My two questions are not showstoppers so we can create new tickets for them is needed. Everything else is all good.

ckanext/ontario_theme/plugin.py
KatiRG: @Alexandra-Green I tested and private files are not shown to users without access. Can you also check?

I believe that this line tells the search to include them (otherwise they will be ignored), but for display on the page it is checked elsewhere if the user has access.

KatiRG:
@Alexandra-Green 

> Does this mean that when a user is sorting by name ascending and clicks a filter, it should then be sorted by last modified?

No, sorry for the confusion. The "Last Modified" sort order should be the default one (i.e. when the user has made no selection) when datasets are listed from any of the four methods:

1. first landing on the Datasets page
2. when clicking on any facet (such as a keyword) to open a list of datasets
3. when clicking on a Ministry in the Ministries page to open a list of datasets for that ministry
4. when clicking on an individual Group

Once a user sorts by something else, that selection should persist throughout.

(At least, this is my understanding from the ticket. The designers will confirm when they review.)

Alexandra-Green:
Private datasets are not shown in search, everything looks good

PR #288, opened 2023-02-22
CKAN 2.9: Header design review updates

Alexandra-Green:
## What this PR accomplishes

Updates the header according to the designers suggestions

## Issue(s) addressed

- Francais link on mobile should remain a full word “Francais“ with no underline
- The Design System suggest having an overlay under the menu drop down to show the hierarchy of the elements. On tablet and mobile the screen is being darkened. 
- Menu button moving to the left and right when pressing it to open and close the drop down

## What needs review

- French link is correctly displaying “Francais“ with no underline 
- When clicking to open the menu, there is an overlay on the page. 
- Transition between the open menu and close menu buttons are seamless.

### DIFF
--- a/ckanext/ontario_theme/fanstatic/common.css
+++ b/ckanext/ontario_theme/fanstatic/common.css
@@ -670,7 +670,7 @@ a:focus {
 .ontario-navigation .ontario-header-button {
   visibility: hidden;
   position: absolute;
-  right: 1rem;
+  right: 0.5rem;
   z-index: 6;
 }
 .ontario-header__menu-toggler {
@@ -794,6 +794,11 @@ the menu dropdowm. Datasets and ministries are displayed in the medium screen na
   }
   .account-text{
     display:none;
+  }
+}
+@media screen and (max-width: 40em) {
+  .ontario-navigation .ontario-header-button {
+    right: 1rem;
   }
 }
 /* ========================================================================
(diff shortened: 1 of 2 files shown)


KatiRG:
> Francais link on mobile should remain a full word “Francais“ with no underline

It has no underline unless you hover over it. Is this ok?

KatiRG:
With menu open on smaller screen, transitioning to larger screen will remove the menu, then if moving back to smaller screen the menu comes back as open. Is this ok or should it not "remember" it's previous open state?

KatiRG:
- [ ] When clicking to open the menu, there is an overlay on the page.
- [ ] Transition between the open menu and close menu buttons are seamless.

Alexandra-Green:
> It has no underline unless you hover over it. Is this ok?

Based on the Figma for hover/active states, yes I believe so
mobile-hover-active

Alexandra-Green:
> With menu open on smaller screen, transitioning to larger screen will remove the menu, then if moving back to smaller screen the menu comes back as open. Is this ok or should it not "remember" it's previous open state?

I feel my expertise in this area falls short, who would be the best person to ask about this? The designers? The Design System team?

KatiRG:
> > With menu open on smaller screen, transitioning to larger screen will remove the menu, then if moving back to smaller screen the menu comes back as open. Is this ok or should it not "remember" it's previous open state?
> 
> I feel my expertise in this area falls short, who would be the best person to ask about this? The designers? The Design System team?

We can ask the designers, and if they want it changed we'll make a ticket

PR #289, opened 2023-02-23
CKAN 2.9 callout tracking 

Alexandra-Green:
## What this PR accomplishes

- Moves the report an error callout to a snippet and adds survey parameters to URL to permit tracking
- Adds survey parameters to the blue feedback callout on the homepage and ministry page

## Issue(s) addressed

Track what page users accessed the form from (e.g., where they were before they clicked on the form) on the report an error callout and blue feedback callout (to match the yellow survey banner at the top of the site)

## What needs review

- Translations are correct for the callouts
- When clicking the survey links in the callouts, for both English and French, the URL will have the page they clicked from as a parameter. If the page they were coming from is a dataset resource page, there should be an id parameter with the resource name.


KatiRG (commented):
> When clicking the survey links in the callouts, for both English and French, the URL will have the page they clicked from as a parameter. If the page they were coming from is a dataset resource page, there should be an id parameter with the resource name.

The `id` in the URL shows the package name, and not the resource name, e.g.:
```
https://www.ontario.ca/form/report-data-error?page=%2Fdataset%2Ffor-local-datastore%2Fresource%2F4eccdeeb-98eb-439a-858a-93b0c4a1b6ed&id=for-local-datastore
```
Here, `for-local-datastore` is the package name, which is correct, and the resource id `eccdee...` is also correct, but the `id` should not also be `for-local-datastore`, but the name of the resource.

KatiRG:
- [ ] Translations are correct for the callouts

Alexandra-Green:
> Here, `for-local-datastore` is the package name, which is correct, and the resource id `eccdee...` is also correct, but the `id` should not also be `for-local-datastore`, but the name of the resource.
I've added the resource name as the id for when users click the link from a datafile page and left the id as the package name when they click from the individual dataset page

ckanext/ontario_theme/plugin.py
```diff
@@ -422,6 +422,20 @@ def extract_package_name(url):
     get_package_name = find_package.findall(url)
     if len(get_package_name) > 0:
         return get_package_name[0]
+    else:
+        return 
+
+def extract_resource_name(url):
```
KatiRG, line 428: This function is almost identical to `extract_package_name()`. Can you one generic function that works for both?

Alexandra-Green: Done!

KatiRG:
looks good!

PR #290, opened 2023-02-28
Ckan 2.9 callout fix

Alexandra-Green:
## Issue(s) addressed

Fixes adding a new resource 502 gateway error caused by callout
Displays resource type if no name is provided and if no name and no type is provided for a resource, "Supporting file" is displayed as the id in the URL

## What needs review

Users can add new resources without a 502 error
Survey URL id parameter displays resource type if there is no resource name entered, but there is a resource type.
Survey URL id parameter displays "Supporting file" when there is no resource type and no resource name.
Survey URL id parameter displays resource name, when name is provided.


KatiRG (approved):
looks good!

PR #292, opened 2023-02-28
Remove image uploader options

KatiRG:
## What this PR accomplishes
Removes the option to upload a profile image for new and existing users, and for new and existing ministries.

## What needs review
- Please verify that the following pages no longer include option to upload an image:
    - existing user profile edit page
    - new user form page 
    - existing ministry edit page
    - new ministry form page
- Check that ministry placeholder image no longer appears on any of the ministry pages
- Confirm that an image can still be uploaded as a resource of a dataset page


Alexandra-Green (commented):
Should the group page image be removed as well?

Alexandra-Green (commented):
✅ Please verify that the following pages no longer include option to upload an image:
        existing user profile edit page
        new user form page
        existing ministry edit page
        new ministry form page
    ✅ Check that ministry placeholder image no longer appears on any of the ministry pages
    ✅ Confirm that an image can still be uploaded as a resource of a dataset page -> this is possible but it's only a link in the image view. Is that expected?

KatiRG:
> ```
> ✅ Confirm that an image can still be uploaded as a resource of a dataset page -> this is possible but it's only a link in the image view. Is that expected?
> ```

Yes, we don't have a tool to display images in the Data Explorer window area AFAIK. Would be a good feature to have though...

PR #295, opened 2023-03-06
Filter cleanup

KatiRG:
## What this PR accomplishes
- reorganizes the order of the filter topics in the Dataset search page according to the figma (Topics, Ministries, Groups, Formats, Licenses, Asset Type, Update Frequency, Access Level) 
- removes all filter options with only 1 choice (e.g., Jurisdiction, etc.)

## What needs review
Please confirm the 2 above changes


Alexandra-Green (commented):
Are the filter options supposed to be hidden if it has one choice after filtering and/or search? or the options that only ever have one option are the ones that will not be displayed?

KatiRG:
I think if it has only one choice _after_ filtering, that is ok.

Alexandra-Green (approved):

PR #298, opened 2023-03-13
Ckan 2.9 Filter modal bug

Alexandra-Green:
## Issue addressed

The filter modal is blocked by the navigation bar and users are incapable of closing the modal without refreshing the page. This problem is only on the dataset search page on screen widths < 765px. (For individual group and Ministry pages the filter results button is too low on the page to cause this problem)

## What needs review

On 765px and smaller screens, the navigation bar should be behind the modal on the dataset search page. The individual group and Ministry pages should be tested as well.


KatiRG (commented):
This bug is not consistently seen across all devices but let's merge !

PR #307, opened 2023-03-20
Fixes count of datasets in Housing, Communities and Social Support

KatiRG:
## What this PR accomplishes
Fixes the count of the number of datasets within the Homepage topic "Housing, Communities and Social Support". Count was previously showing as 0 because the function that matches the Homepage topic string with the keyword string did not disregard the comma. (Keyword strings in packages are not allowed to contain commas.)

## What needs review
Please confirm that the number of datasets in the Homepage topic "Housing, Communities and Social Support" is now showing the correct count.

### DIFF
--- a/ckanext/ontario_theme/plugin.py
+++ b/ckanext/ontario_theme/plugin.py
@@ -381,14 +381,32 @@ def get_group_datasets(group_id):
 
 def get_keyword_count(keyword_lang, language):
     '''Helper to return the dataset count of the indicated group by language
+
+    keyword_lang
+        The topic category string on the Home page.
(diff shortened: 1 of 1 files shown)


Alexandra-Green:
This is working for me for Housing Communities and Social Support, this PR isn't taking into account that Culture and Recreation is not working on Stage?

KatiRG:
Not sure what was happening with Culture and Recreation. There was something weird with the Ministry name overlapping with another one, so I cleaned it up and the topic count works now:
Screen Shot 2023-03-20 at 21 07 26

Alexandra-Green (approved):

PR #312, opened 2023-03-21
Display resource page content only if dataset is open or user admin

aimalkhan:
## What this PR accomplishes
* Only allow admin/sysadmins access to data that is not open. Users not having these permissions or that are not logged in are not shown resource information.
## Issue(s) addressed
* Users not logged in can access under-review/restricted URLs and view the resource contents
## What needs review
* if-else condition placed in `{% block pre_primary %}`
* if condition placed in `{% block primary %}`
## What needs to be tested
* open/under-review/restricted resource pages load for logged in user.
* open resource pages load for not-logged in user.
* under-review/restricted resource url when accessed by a not-logged in user displays "Resource not available", and no resource related info is leaked.
image

### DIFF
--- a/ckanext/ontario_theme/templates/internal/scheming/package/resource_read.html
+++ b/ckanext/ontario_theme/templates/internal/scheming/package/resource_read.html
@@ -15,221 +15,231 @@
 {% set license = h.ontario_theme_get_license(pkg.license_id) %}
 
 {% block pre_primary %}
-  <section class="module module-resource" role="complementary">
-    {% if h.check_access('package_update', {'id':pkg.id }) %}
-      <header class="col-xs-12 page-header">
-      <div class="content_action">
-        {% link_for _('Edit'), named_route=pkg.type + '_resource.edit', id=pkg.name, resource_id=res.id, class_='btn btn-default content_action', icon='wrench' %}
-      </div>
-      </header>
-    {% endif %}
-    {% block resource_content %}
-      <div class="module-content">
-        <div class="col-lg-8 col-md-7 col-xs-12">
-          {% block resource_read_title %}
-            <h1 class="page-heading">{{ h.resource_display_name(res) }}</h1>
-          {% endblock %}
-          <div class="datafile-p" property="rdfs:label">
-            {% block resource_read_url %}
-              {% if res.url and h.is_url(res.url) %}
-                <p>{{ _('URL:') }} <a class="resource-url-analytics" href="{{ res.url }}" title="{{ res.url }}">{{ res.url }}</a></p>
-              {% elif res.url %}
-                <p>{{ _('URL:') }} {{ res.url }}</p>
-              {% endif %}
-            {% endblock %}
-            {% if res.description %}
-              {{ h.render_markdown(h.get_translated(res, "description")) }}
-            {% endif %}
-            {% if not res.description and c.package.notes %}
-              {% if res.datastore_active %}
-                {{ h.render_markdown(h.get_translated(c.package, 'notes')) }}
-              {% else %}
-                <h3>{{ _('From the dataset abstract') }}</h3>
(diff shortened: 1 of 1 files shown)


ckanext/ontario_theme/templates/internal/scheming/package/resource_read.html
```diff
@@ -15,221 +15,233 @@
...
-          {% block resource_view_content %}
-            {{ super() }}
-          {% endblock %}
-        {% endif %}
-      {% endblock %}
-    </div>
-  </section>
+    </section>
+  {% else %}
+    <h2>{{ _('Resource not available') }}</h2>
```
Alexandra-Green, line 151:
```suggestion
    <h1 class="col-xs-12">{{ _('Resource not available') }}</h1>
```
Since it's the only heading in the page it should be an `<h1>` and the added class is to align the left margin on all screens.

Alexandra-Green (changes requested):
The text is not translated in French.
For "Data Not Available" there is already a translation, if you wanted to use that text instead. 

✅ I am able to see all datasets when I'm logged in as an admin
✅ I am unable to use the link of a restricted or under review dataset as a non admin user to view a dataset. I am shown "Resource not available" and no resource related information is linked

KatiRG:
- [ ] open/under-review/restricted resource pages load for logged in user.
- [ ] open resource pages load for not-logged in user.
- [ ] under-review/restricted resource url when accessed by a not-logged in user displays "Resource not available", and no resource related info is leaked.

KatiRG:
You could use the same message as is posted on the dataset page (either copy the whole snippet or just take the text):
Screen Shot 2023-03-21 at 13 33 37

and the French:
Screen Shot 2023-03-21 at 13 33 59

ckanext/ontario_theme/templates/internal/scheming/package/resource_read.html
```diff
@@ -15,221 +15,233 @@
...
-                <p>{{ _('URL:') }} <a class="resource-url-analytics" href="{{ res.url }}" title="{{ res.url }}">{{ res.url }}</a></p>
-              {% elif res.url %}
-                <p>{{ _('URL:') }} {{ res.url }}</p>
-              {% endif %}
+  {# only display resource page content if open dataset or admin user #}
+  {% set is_admin = h.check_access('package_update', {'id':pkg.id }) %}
+  {% set is_open = (pkg.access_level == 'open') %}
+  {% if is_open or is_admin%}
+    <section class="module module-resource" role="complementary">
+      {% if h.check_access('package_update', {'id':pkg.id }) %}
```
KatiRG, line 23: Do you need to check this again? `is_admin` and `is_open` are already satisfied

ckanext/ontario_theme/templates/internal/scheming/package/resource_read.html
aimalkhan: i will make this and the translation change.

ckanext/ontario_theme/templates/internal/scheming/package/resource_read.html
aimalkhan: the variable definition did not carry over to the different block, so had to redefine them.
I will use "Data Not Available" / "Données non disponibles"

KatiRG: I meant, there is already an if condition on line 21, why do you need to check again on line 23?

aimalkhan: oh, right.
23 is from code already in place, and I was not sure whether to remove it or not. it is the only repeated-check happening. Removing it will simplify the logic, so i will make the change.

Alexandra-Green (approved):
Looks great

PR #315, opened 2023-03-23
Translate about page

KatiRG:
## What this PR accomplishes
Dirty hack to get the about page to translate to french in CKAN 2.9

## What needs review
- confirm that the page translates
- if you want to find a better way than the ridiculous `gettext()` chunks, feel free!

PR #316, opened 2023-03-23
Alternate way of translating the about page

Alexandra-Green:
## What this PR accomplishes

Alternate way of translating the about page for CKAN 2.9 to #315

## What needs review

- The page translates

## Alternate method to translate

- Recreate the .pot file using `python setup.py extract_messages`
- Copy the extracted translation for the about page.
- Revert to the old .pot file and paste what has been copied in the `msgid` for the about page in the `.po` and old `.pot`  file
- Compare the content changes in the about page to the about page translation

**Do not commit the new extracted `.pot` file, it will delete some translations. It should only be used to copy the extracted about page HTML. There is probably a way to only extract that page, I will look into it**


KatiRG:
> ## Alternate method to translate
> 
>     * Recreate the .pot file using `python setup.py extract_messages`
> 
>     * Copy the extracted translation for the about page.
> 
>     * Revert to the old .pot file and paste what has been copied in the `msgid` for the about page in the `.po` and old `.pot`  file
> 
>     * Compare the content changes in the about page to the about page translation

also need to recompile: `python3 setup.py compile_catalog`

KatiRG (approved):
Yes this is a better method! Less error-prone, even though some manual gymnastics still required.

PR #317, opened 2023-03-23
Replaced all archived OD Directive links to Digital and Data Directive

rleblanc-on:
### What this PR accomplishes
An Open Data lead recently [noted](https://ontariodigital.atlassian.net/browse/DATAMAIL-109) that some links on the catalogue were directing to an archived directive, the Open Data Directive, which has been replaced by the [Digital and Data Directive](https://www.ontario.ca/page/ontarios-digital-and-data-directive-2021)
All text on the catalogue that referenced this previous directive and its URL (in English and French) has been replaced.


ckanext/ontario_theme/templates/external/package/snippets/ontario_theme_access_level.html
```diff
@@ -9,7 +9,7 @@
       <strong>{{ _("Data Available") }}</strong>
       <p>
         {% trans license_url=_(pkg.license_url) %}
-        The data described here is available for you to use. <a href="https://www.ontario.ca/page/ontarios-open-data-directive">[Learn more]</a><br />
+        The data described here is available for you to use. <a href="https://www.ontario.ca/page/ontarios-digital-and-data-directive-2021">[Learn more]</a><br />
```
Alexandra-Green, line 12: Can you define the URL as a variable at the beginning of the html and add it into the trans tags similar to license_url? That way if the link ever changes again we only need to change it in two places instead of however many times you had to change it this time. 

You'd be able to translate just the url in the .pot and .po files and replace whats after the `href=\"` in these files with %(<variable_name>)s.

Alexandra-Green (changes requested):

ckanext/ontario_theme/templates/external/package/snippets/ontario_theme_access_level.html
rleblanc-on: @Alexandra-Green ready for review. 😀

Alexandra-Green (changes requested):

Alexandra-Green (approved):

aimalkhan (approved):
👍 
seems like there is still a merge conflict with the `ckanext-ontario_theme.mo` file, and it needs a rebase.

PR #318, opened 2023-03-24
prevent orgs from being sorted on initial launch

KatiRG:
## What this PR accomplishes
Prevents the function that sorts organization names from being called when there are no organizations in the application.

## What needs review

- Check on the pre-launch version that the Ministries tab does not crash.
- Check on local version that Ministry sort and search still works the same


rleblanc-on:
- [x] Check on the pre-launch version that the Ministries tab does not crash
- [x] Check on local version that Ministry sort and search still works the same

PR #319, opened 2023-03-29
Fix broken group page

KatiRG:
## What this PR accomplishes
Fixes the broken group page by adding a check to use group title if no titles are translated.

## What needs review
Confirm that group page displays when a group does not have a `title_translated` field.


aimalkhan (approved):
looks good! 👍

Alexandra-Green (approved):
Looks good to me!

PR #320, opened 2023-03-31
make Relevance the default sort order

KatiRG:
## What this PR accomplishes
Sets the default sort order of datasets to Relevance instead of Last Modified.

## What needs review
Check that Relevance is the default sort order of datasets when:

1. First landing on the Dataset search page
2. Opening a dataset from the Ministry search page
3. Selecting a Group from the Groups page

Also check:
- if the user selects own sort order, that sort order is preserved after interactions such as using the search bar or paginating through the list of datasets
- equivalent functionality of all of the above in French

### DIFF
--- a/ckanext/ontario_theme/templates/internal/group/read.html
+++ b/ckanext/ontario_theme/templates/internal/group/read.html
@@ -23,9 +23,9 @@ block. super() will end up calling the facets twice.
     (_('Last Modified'), 'metadata_modified desc'),
     (_('Popular'), 'views_recent desc') if g.tracking_enabled else (false, false) ]
   %}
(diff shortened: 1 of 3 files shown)


Alexandra-Green (approved):
Check that Relevance is the default sort order of datasets when:

-     First landing on the Dataset search page ✅ 
-     Opening a dataset from the Ministry search page ✅ 
-     Selecting a Group from the Groups page ✅ 
-     if the user selects own sort order, that sort order is preserved after interactions such as using the search bar or paginating through the list of datasets ✅ 
-     equivalent functionality of all of the above in French ✅

PR #333, opened 2023-05-05
DATA-654 Group names display and translation

Alexandra-Green:
## What this PR accomplishes

- Displays and translates the group's title in the breadcrumbs of the individual groups pages

## Issue(s) addressed

- Group names were not being translated in the breadcrumbs.

## What needs review

- The breadcrumb on an individual group page displays the group's title and not the URL's name for the page. 
    - (In the url name, there are no capitals and spaces are replaced with hyphens)
    - (The group's title is displayed on the left side, above the filters. This should be used as reference for the breadcrumb for the page)

## Note:

On local, test and stage the breadcrumb is different than the prod breadcrumb for the groups individual page. The breadcrumb for the group page does not translate either for any of the environments. This PR is to translate the breadcrumb, but it also fixes the way the breadcrumbs look on local, stage and test.


KatiRG (approved):

KatiRG:
works as expected on local, test, and stage

PR #334, opened 2023-05-05
Updates relevant Design System modules

Alexandra-Green:
## What this PR accomplishes

- Updates relevant Design System modules from their 1.2.1 release

## Issue(s) addressed

- aria-label in header search for mobile, and missing role in the Ontario logo

## What needs review

- The header search toggler has the aria-label "Search" in tablet and mobile.
- Ontario logo in top left has the `role="img"`


KatiRG:
> The header search toggler has the aria-label "Search" in tablet and mobile.

confirmed. This aria-label is also in full-size screens. Is that fine?

> Ontario logo in top left has the role="img"

confirmed

PR #336, opened 2023-05-09
fix api modal layout for safari by adding min-width

KatiRG:
## What this PR accomplishes
Fixes improper formatting of the API modal in Safari browser.

## What needs review
- Open the API modal in Safari browser and verify that the formatting is ok.
- Check that the fix does not mess up the API modal in other browsers


Alexandra-Green (approved):
Mobile in safari looks good & other browsers are not affected ✅

PR #338, opened 2023-05-10
Restricted under review datasets

Alexandra-Green:

- Adds "Not available to the public" for Restricted/Under Review datasets
- Removes the resource format types for Restricted/Under review datasets

## What needs review

- Disclaimer for restricted and under review datasets matches figma and are applied to the correct access level
- Translations are correct
- Resource format types are not visible to non-admin users
- Admin users are able to see the resource format types along with the disclaimer (I added this as a security precaution), determine if this is needed/wanted. 
- Descriptions are not visible in the dataset previews
- Determine if it is needed/wanted for admin users to be able to see the descriptions of the datasets in the previews


KatiRG:
> * Disclaimer for restricted and under review datasets matches figma and are applied to the correct access level

confirmed
> 
> * Translations are correct

need french file

> 
> * Resource format types are not visible to non-admin users

confirmed

> 
> * Admin users are able to see the resource format types along with the disclaimer (I added this as a security precaution), determine if this is needed/wanted.

confirmed. Designers to decide in QA

> 
> * Descriptions are not visible in the dataset previews

confirmed

> 
> * Determine if it is needed/wanted for admin users to be able to see the descriptions of the datasets in the previews

Designers to decide in QA

KatiRG (approved):

PR #340, opened 2023-05-11
Fix vulnerable JavaScript dependency

Alexandra-Green:
## What this PR accomplishes

- Removes the webservices.gov.on.ca script from the code

## Issue(s) addressed

- Using components with unpatched vulnerabilities can increase the risk profile of the application and can be exploited by attackers under certain conditions

## What needs review
- On local, there should be no reference of a webservices.gov.on.ca script
- Test / Stage must also be reviewed as there is a second script that will not run on localhost. Confirm there are no references of webservices.gov.on.ca js file or php file on Stage and/or Test.

### DIFF
--- a/ckanext/ontario_theme/templates/internal/base.html
+++ b/ckanext/ontario_theme/templates/internal/base.html
@@ -2,9 +2,6 @@
 
 {%- block links -%}
   {{ super() }}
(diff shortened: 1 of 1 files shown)


aimalkhan (approved):
✅ No reference to metrics.js file or metrics.php file on test

PR #342, opened 2023-05-15
Displays group name as subtitle

Alexandra-Green:

- Group name displayed in tab was displaying URL title instead of the group's title name
Screenshot from 2023-05-15 15-14-11
Look at the tab compared to group title

## What needs review

- Name displayed in tab is the same as breadcrumb and group title and translates to French


Alexandra-Green:
> @Alexandra-Green Please change group name in https://github.com/ongov/ckanext-ontario_theme/blob/Group_page_names/ckanext/ontario_theme/templates/internal/home/snippets/ontario_theme_covid_datasets.html#L6 to new name. Also line 20 https://github.com/ongov/ckanext-ontario_theme/blob/Group_page_names/ckanext/ontario_theme/templates/internal/home/snippets/ontario_theme_covid_datasets.html#L20
> 
> perhaps better to store as a variable?

Done!

KatiRG (approved):

PR #344, opened 2023-05-16
Corrects "Asset type" filters that do not translate properly to French

Alexandra-Green:
## What this PR accomplishes

- Corrects "Asset Type" translation in filters

## Issue(s) addressed

- The two asset type filter categories (Dataset and Algorithm) currently translate in french incorrectly
- Asset Type display name was being incorrectly translated

## What needs review

- The "Asset Type" Filter in the dataset search page, individual group pages, and individual ministry page should now be:

Asset type = Types d’actifs

Dataset = Jeu de données

Algorithm = Algorithme


KatiRG (approved):
> 
>     The "Asset Type" Filter in the dataset search page, individual group pages, and individual ministry page should now be:
> -   Asset type = Types d’actifs
> -   Dataset = Jeu de données
> -   Algorithm = Algorithme

confirmed

PR #347, opened 2023-05-17
Logout button fails

Alexandra-Green:
## What this PR accomplishes

- Redirects users to the logout page after pressing the logout button

## Issue(s) addressed

- Logout button erroneously had a '_' at the beginning of the logout html file in the relative path, causing issues with actually logging users out when the button was pressed.

## What needs review

- On test or stage (because local works fine), users should be redirected to the logout page after clicking the logout button. 
- Hovering over the logout button, the path should end with `user/logout` (no '_')


aimalkhan (approved):
👍

PR #352, opened 2023-05-24
add checks to resource name and type

KatiRG:
## What this PR accomplishes
Places checks on `resource_name` attributes in function `extract_package_name()` that was causing errors in the `uwsgi.ERR` log file.

## What needs review
To make sure the code checks for all possibilities, upload new resources with different combinations of name and type and make sure there are no errors in `uwsgi.ERR` or on the page:
- no name, no type 
- no name, but type Data
- no name, but type Information
- add a name, but no type
- add a name and a type (try different types)

Test the above for uploaded CSV resources and for link URLs.

### DIFF
--- a/ckanext/ontario_theme/plugin.py
+++ b/ckanext/ontario_theme/plugin.py
@@ -678,11 +678,13 @@ def extract_package_name(url):
             resource_name = toolkit.get_action('resource_show') (
                 data_dict={'id': get_resource_name[0]}
                 )
-            resource_type = resource_name['type']
-            resource_name = resource_name['name']
-            if not resource_type and not resource_name:
-                resource_name = "Supporting File"
-            return resource_name or resource_type
+            if not resource_name['type'] and not resource_name['name']:
+                return "Unnamed Supporting File"
+            if 'name' in resource_name:
+                if len(resource_name['name']) > 0:
+                    return resource_name['name']
+                else:
+                    return "Unnamed Data File"
         except ckan.logic.NotFound:
             return False
     elif len(get_package_name) > 0:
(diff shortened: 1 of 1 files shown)


ckanext/ontario_theme/plugin.py
```diff
@@ -678,11 +678,13 @@ def extract_package_name(url):
...
-            if not resource_type and not resource_name:
-                resource_name = "Supporting File"
-            return resource_name or resource_type
+            if not resource_name['type'] and not resource_name['name']:
+                return "Unnamed Supporting File"
+            if 'name' in resource_name:
+                if len(resource_name['name']) > 0:
+                    return resource_name['name']
+                else:
+                    return "Unnamed Data File"
```
Alexandra-Green, line 687: Why did you opt to using "unnamed data file"? The "id" parameter will now lose some of the context `resource_name['type']` provided since resources with a type and no name are given the type as the name

KatiRG: What should it return instead?

KatiRG:
> This solution does not present any errors on my local, but I haven't found a resource where `resource_type` is not null. Can you find an example? In the case `resource_type` is never not null, the checks would only have to be done on `type`, right?

This is precautionary and possibly overkill, but better to have a check than find some example later where it fails. I have found a case where `resource_name` has no `type`, so just erring on the side of caution.

Alexandra-Green:
> This is precautionary and possibly overkill, but better to have a check than find some example later where it fails. I have found a case where `resource_name` has no `type`, so just erring on the side of caution.

Sounds good!

Alexandra-Green (approved):

PR #360, opened 2023-06-05
Search accurary and character tolerance

azimhamza:
### What this PR accomplishes & Issue(s) addressed

- aims to enhance the search functionality by addressing issues of limited search functionality due to standard tokenization and errors resulting from special characters in the search queries. 

1. **N-gram Tokenization**: Implemented n-gram tokenization on 'name' and 'title' fields to improve partial word queries and queries with slight misspellings. Created new fields 'name_ngram' and 'title_ngram' for storing the n-gram tokenized versions of 'name' and 'title' fields respectively to preserve the integrity of the original data. 

2. **Character Escaping**: Introduced Solr's solr.String field type for fields handling special characters, which automatically escapes special characters, thereby preventing errors in SOLR search caused by special characters.

### Issue(s) Addressed

- Limited Search Functionality (Data 870, Data 873)
- Special Characters Causing Errors in SOLR Search (Data 872)

### What needs review

The new n-gram tokenization and character escaping implementations need to be tested thoroughly. The primary areas of focus should be:

- Search functionality with partial word queries or slight misspellings.
- Search functionality when users include special characters in their search queries.
- Ensuring no errors occur due to special characters.
- Performance impact due to n-gram tokenization.

It is important to confirm that these changes have improved the search experience and returned more accurate results, while the performance of the search remains optimal.


aimalkhan:
why was `text_ngram` removed?

PR #361, opened 2023-06-05
change group name in group url in homepage

KatiRG:
## What this PR accomplishes
Changes the group name for the homepage link for covid to reflect the new name.

## What needs review
The group name in Stage and Test has been changed to the new name. Test that this fix ensures that covid datasets are linked to the correct group page from the home page.


KatiRG:
already done in another PR!

PR #362, opened 2023-06-08
Homepage topic links

Alexandra-Green:
## What this PR accomplishes

- Put topics into a list
- Moved the numbers outside the a tag, to remove conflict with GA4

## Issues addressed

- The number of datasets shouldn’t be a part of the URL because it breaks the GA statistics. Move the count outside of the <a href tag.

## What needs review
- Numbers are not in the href
- Clicking the image or text, the blue focus border is around both the text and the image (not the number)
- Hovering over the text or image will change the colour of the image and underline the text (not the number)
- Numbers do not appear in GA4

### DIFF
--- a/ckanext/ontario_theme/fanstatic/common.css
+++ b/ckanext/ontario_theme/fanstatic/common.css
@@ -2219,3 +2219,32 @@ screen and (min-width: 96em) {
 .module.module-narrow.module-shallow.context-info {
   padding: 30px;
 }
+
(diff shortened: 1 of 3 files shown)


KatiRG:
- [ ]     Numbers are not in the href
- [ ]     Clicking the image or text, the blue focus border is around both the text and the image (not the number)
- [ ]     Hovering over the text or image will change the colour of the image and underline the text (not the number)
- [ ]     Numbers do not appear in GA4

PR #367, opened 2023-06-14
Ministry about page translations

Alexandra-Green:
## What this PR accomplishes

- Filters the dictionary that has the value of the selected Jurisdiction/Category option and returns the object of only labels
- Using the environment language variable, filters for the appropriate language label and prints

## Issue(s) addressed

- Juridiction and Catégorie content does not translate

## What needs review

- Content is now translated on the ministry about page


KatiRG (commented):
You can eliminate the inner `if` statement by pre-filtering out the excluded fields already in the opening `for` loop:

```
{% set about_fields = c.scheming_fields|rejectattr("field_name", "in", exclude_fields ) | list  %}
<dl class="inline-dl">
    {% for f in about_fields %}
        <dt>{{ h.scheming_language_text(f.label) }}{{ _(':') }}</dt>
        <dd>
          {% set translated_labels = f.choices|selectattr("value", "equalto", c.group_dict[f.field_name] )|map(attribute='label') %}
          {{ translated_labels|map(attribute=current_lang)|join }}
        </dd>
    {% endfor %}
</dl>
```

KatiRG (approved):

PR #372, opened 2023-06-27
Facet UI changes

Alexandra-Green:
## What this PR accomplishes

- Reorders facets across site
- Changes Show more link into button
- Remove facets with one element for individual Ministry page and Group page
- Creates snippet for facet rearrangement

## Issue(s) addressed

- Orders facets by priority
  - Topics
  - Ministries
  - Formats 
  - Access level
  - Update frequency
  - Licences
  - Asset type
  - Groups
- Show more link is a secondary button
- Removes jurisdiction and category facets if there is only one element.
- Updates styling to match figma
- Adds translation to jurisdiction and category facets

## What needs review

- If there is more than one jurisdiction or category, it appears in the facets list, else it is not in the facet column.
- If a jurisdiction or category facet option is active, the facet option is still visible to the user and they can remove the option from the facet box
- Show more button functions the same as the show more link
- Facets are in the above order on all pages.
- Facets are responsive
- Jurisdiction and category facets translate
- Styling matches Figma, except:
   - facet column is 25% of total width and packages are 75%
   - No ellipses between show more and the last facet option.


KatiRG:
> If there is more than one jurisdiction or category, it appears in the facets list, else it is not in the facet column

I tested by changing a Ministry to be a Federal jurisdiction and having category Agency:

Screenshot 2023-07-06 at 15 14 50

This should add a second option to both the Jurisdiction and Category facets and these should appear in the facet column, but they are not there in the Datasets search page:
Screenshot 2023-07-06 at 15 16 47

KatiRG:
- [ ]     Orders facets by priority
        Topics
        Ministries
        Formats
        Access level
        Update frequency
        Licences
        Asset type
        Groups
- [ ]     Show more link is a secondary button

KatiRG:
- [ ] If there is more than one jurisdiction or category, it appears in the facets list, else it is not in the facet column.
- [ ] If a jurisdiction or category facet option is active, the facet option is still visible to the user and they can remove the option from the facet box
- [ ] Jurisdiction and category facets translate

KatiRG (approved):

PR #378, opened 2023-07-11
Fixes resource page showing "Data not available" instead of data table

Alexandra-Green:
## Issue(s) addressed

- Fixes bug causing the resource page to only show "Data not available" for open datasets
- Fixes bug causing the edit resource button to not appear on resource page
- Bugs introduced by right hand styling PR

## What needs review

- All users can see the data table and right hand boxes on open datasets
- Non admin users only see "Data not available" when trying to access under review or restricted resources. 
- Admin users can edit resources


ckanext/ontario_theme/templates/internal/scheming/package/resource_read.html
```diff
@@ -19,143 +19,152 @@
 {% set license = h.ontario_theme_get_license(pkg.license_id) %}
 {% set is_admin = h.check_access('package_update', {'id':pkg.id }) %}
 {% set is_open = (pkg.access_level == 'open') %}
```
KatiRG, line 21: `is_open` is defined globally here but is not available to the blocks, so can be removed here.

KatiRG (changes requested):
Some format fixes are required on this file (`djlint ckanext/ontario_theme/templates/internal/scheming/package/resource_read.html --check`).

KatiRG (changes requested):
A non-admin user can still see dataset page descriptions for datasets that are marked as Under Review or Restricted, instead of "Data not available".

Alexandra-Green:
> Some format fixes are required on this file (`djlint ckanext/ontario_theme/templates/internal/scheming/package/resource_read.html --check`).

Okay so I made some of the changes, but I had to revert the ones in the {% block primary %} because it was adding extra words and causing errors, but everything should be formatted correctly, I'll be working on the linter ticket to get the formatter working correctly.

KatiRG (approved):

PR #379, opened 2023-07-12
Download/about/preview buttons on Dataset page

Alexandra-Green:
## What this PR accomplishes

- Adds 2px white label border to language label (Ticket asked for 1px but it was not making a difference)
- remove file sizes from buttons and adds the size to the file name 
- remove languages from buttons
- set min width of 8 characters + icon (e.g. the “Download” button) 
- set min height to 1 line of text
- Adds French translations and abbreviations to file sizes
- Replaces CKAN core grid system with DS grid system in the resource item list

## Issue(s) addressed

- DATA-1017 (1px border)
- DATA-1121 (French File Sizes)
- DATA-935
- On mouse hover the language label almost blends in with the gray background.  It starts looking like a part of the clickable text but it is not clickable. It could be confusing for our users 
- We currently use the prefixes in place of symbols for file sizes on the French catalogue. This is incorrect and doesn’t actually specify a unit of measurement.

## What needs review

- File sizes have abbreviations and translate accordingly. File sizes also appear in additional information table on resource page
- Buttons match design outline in comments of DATA-935
- White label border around language label is visible.
- Resource list is responsive and changing grid systems does not cause an overflow or misalignment.


aimalkhan (approved):
👍 
needs a rebase.

aimalkhan (approved):
tested on ist.
I do not have a better suggestion than implementing `def abbr_localised_filesize`, and will approve the PR.

PR #380, opened 2023-07-14
Fixes internal 500 error when using load_organization_data.py

Alexandra-Green:
## What this PR accomplishes & issue(s) addressed

- When running the load_organizations_data.py, all resources being loaded will get an internal 500 error. 
- This PR includes a minor fix and a check for packages with resources

## What needs review

- Running load_organization_data.py, you do not get an internal 500 error
- Add a resource, input a random/unique lowercase file format. On the dataset page, in the formats facet box on the dataset search page, and in the api `/api/3/action/resource_show?id=<id>`, the format is in all uppercase.

### DIFF
--- a/ckanext/ontario_theme/resource_upload.py
+++ b/ckanext/ontario_theme/resource_upload.py
@@ -168,5 +168,6 @@ class ResourceUpload(DefaultResourceUpload):
         if url and not (resource.get('url_type') == 'upload') and not resource.get('format'):
             resource['format'] = 'WEB'
 
-        if not (resource.get('format') == 'GeoJSON'):
-            resource['format'] = resource['format'].upper()
+        if resource.get('format') and not (resource.get('format') == 'GeoJSON'):
+            resource_format = resource.get('format')
+            resource['format'] = resource_format.upper()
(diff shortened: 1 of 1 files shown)


ckanext/ontario_theme/resource_upload.py
```diff
@@ -168,5 +168,6 @@ def _check_zip_mimetype(self):
         if url and not (resource.get('url_type') == 'upload') and not resource.get('format'):
             resource['format'] = 'WEB'
 
-        if not (resource.get('format') == 'GeoJSON'):
-            resource['format'] = resource['format'].upper()
\ No newline at end of file
+        if resource.get('format') and not (resource.get('format') == 'GeoJSON'):
+            resource_format = resource.get('format')
+            resource['format'] = resource_format.upper()
```
aimalkhan, line 173: why not use 
```resource['format'] = resource.get('format').upper()```

 if `resource_format` is not being used elsewhere?

aimalkhan (changes requested):

ckanext/ontario_theme/resource_upload.py
Alexandra-Green: Done

aimalkhan (approved):
looks good; and the uppercase fix also works.

PR #226, opened 2022-07-29
Script to install ckan locally in dev and prod modes

aimalkhan:
implements DATA-866
* add scripts to install ckan on a local VM in both developer and production modes
* add supporting configuration files for local setup
* add supporting documentation for local setup

PR #388, opened 2023-08-01
Fix branch switch with local ontario_theme setup

aimalkhan:
## What this PR accomplishes
* fixes issues caused by merging of #226 into main with branching

## Issue addressed
* fix branch switch to non-existent branch when setting up local ontario_theme

## What needs testing & review
* local script setup works when working directly from main branch


aimalkhan:
tested on local, and it works.

Alexandra-Green (approved):

PR #402, opened 2023-08-30
Task: Update dataset page with new single CTA button and file/format types content

Alexandra-Green:
## What this PR accomplishes

- Changes learn more file types is changed to formats and translated
- Removes download button, replaces it with one CTA button "Explore"
- Button is translated
- For non-admin users entire list item is clickable.

## Issue(s) addressed

- Call to action for each resource is a single, secondary button with updated label 
- Clicking on item opens individual resource page for the respective resource
- Content below resource header (above list of resources) is updated with new content
- The look and feels matches the figma

## What needs review

- Learn more about accessing data using different **formats**. This translates correctly in French
- Explore button translates and links to the resource page

For non-admin users:
- Whole list item is clickable 
- Dataset title is underlined on list item hover
- Button is in the Hover state on list item hover


KatiRG:
- [ ]  Learn more about accessing data using different formats. This translates correctly in French
- [ ]     Explore button translates and links to the resource page

For non-admin users:

- [ ]   Whole list item is clickable
- [ ]     Dataset title is underlined on list item hover
- [ ]     Button is in the Hover state on list item hover

KatiRG:
As an admin user, the Explore button is left white when the whole div is hovered over, but when not logged in, the Explore button changes colour.

KatiRG (changes requested):
only the one comment, otherwise looks good!

Alexandra-Green:
This is intentional. Because admin users have two buttons to choose from, the edit and the explore, the changes requested in the ticket are only viable for public users. If admin users are also required to have only the explore button and no edit button on the dataset resource list, then this change can be done for admin users.

KatiRG (approved):

PR #403, opened 2023-08-30
Task: use table view for Groups page

Alexandra-Green:
## What this PR accomplishes

- Puts Groups search page into HTML div table view
- Adds survey to bottom of page
- Translates column headers

## Issue(s) addressed

- DATA-1050
- Completes Easy Wins V1 -> can now be marked as complete rather than work in progress.

## What needs review

- in Test or Stage, screen reader must be used to confirm layout is accessible
- Column headers translates
- Table layout matches Figma
- Links work


KatiRG:
- [ ] Column headers translates
- [ ] Table layout matches Figma
- [ ] Links work

KatiRG:
> in Test or Stage, screen reader must be used to confirm layout is accessible

Screen reader (Chrome extension) reads the group names but not:
- the table header names (Groups, Datasets)
- the number of datasets in each group

KatiRG (changes requested):
Please confirm if screen reader should be reading table column header and the number of datasets in each group.

Alexandra-Green:
Using NVDA, the screen reader reads out the column headers and the amount of dataset each group has. I also tried with the chrome extension screen reader, and it reads everything.

KatiRG (approved):

PR #407, opened 2023-09-15
[Task]: Show more button functionality

Alexandra-Green:
## What this PR accomplishes

- Switched show more links to buttons
- Show more button opens without redirecting to a new page.
- On click of show more button, show more button is hidden and show less button is shown & vice versa.

## Issue(s) addressed

- When a user clicks a show more or show less button, without redirecting to a new page, all the facet options are visible or hidden (depending on the desired action) to the user

## What needs review

- On click of show more button, show more button is hidden and show less button is shown & vice versa.
- Clicking on facet options work as expected.


Alexandra-Green:
@KatiRG

> in the ckan.ini search.default.facets needs to be set to 50 for this to function properly

Did you set the search.default.facets to 50?

KatiRG:
> @KatiRG
> 
> > in the ckan.ini search.default.facets needs to be set to 50 for this to function properly
> 
> Did you set the search.default.facets to 50?

sorry forgot! it works now so nvm

KatiRG:
- [ ]     On click of show more button, show more button is hidden and show less button is shown & vice versa.
- [ ]     Clicking on facet options work as expected
- [ ]     Works in EN and FR

KatiRG (changes requested):
I recommend reworking this a bit to use the limit `search.facets.default` as it is [defined](https://docs.ckan.org/en/2.9/maintaining/configuration.html#search-facets-default), which is the number of facets to show, to avoid confusion. Otherwise it ends up getting overridden in your code as a hard-coded limit of 5. 

I would look at how this limit is implemented in the vanilla ckan code without the ontario theme, because there is a mixup somewhere: if `search.facets.default` is 6, for example, then in the vanilla ckan, 6 facets items are shown for each category, unless the Show more link is clicked, then all items are shown. In the theme, however, 5 items are always shown, and the "Show more" will **show only 6 items** for all facets except for Topics and Ministries, for which all items are correctly displayed.

I'll make some in-line comments to flag a few things.

ckanext/ontario_theme/templates/internal/snippets/facet_list.html
```diff
@@ -47,15 +48,14 @@ <h4 class="ontario-h5">
 
       <p class="module-footer">
         {% if h.get_param_int('_%s_limit' % name) %}
-          {% if h.has_more_facets(name, c.search_facets) %}
-            <a href="{{ h.remove_url_param('_%s_limit' % name, replace=0, extras=extras, alternative_url=alternative_url) }}"
-               role="button"
-               class="facets-show-more ontario-button ontario-button--tertiary">{{ _('Show more') }}</a>
+          {% if items|length > 5 %}
```
KatiRG, line 51: Reinstate the original condition `{% if h.has_more_facets(name, search_facets or c.search_facets) %}` to use the limit as specified by `search.facets.default`, rather than hard-coding `5` here.

KatiRG: You will have to figure out why line 26 `{% with items = items or h.get_facet_items_dict(name) %}` doesn't return all items, only the number defined in `search.facets.default`.

Alexandra-Green:
@KatiRG The `search.facets.default` is used for the "Default number of facets shown in search results", ckan core uses the show more as a link and in the query parameters makes the limit of facets shown be 0, see [https://data.ontario.ca/dataset/?_keywords_en_limit=0](https://data.ontario.ca/dataset/?_keywords_en_limit=0). With this link `search.facets.limit` is being used which is defaulted to 50.

With CKAN core, the default 5 facet options will live on the page. The user clicks show more, they are directed to a link that has the max 50 facet options living on the page.

So with the scenario you described with the 6 show more items, it is because only those 6 items live on the page due to the defined `search.facets.default`.

KatiRG:
@Alexandra-Green that's correct, the number defined in `search.facets.default` should reflect the number of facet items displayed under each facet category. if we define `search.facets.default` as 50, we are not going to show 50, we only show 5 which is a hard-coded limit.

I am suggesting to use the limit of 5 in `search.facets.default`, which is consistent with its definition.

ckanext/ontario_theme/templates/internal/snippets/facet_list.html
```diff
@@ -34,7 +34,8 @@ <h4 class="ontario-h5">
             {% set label_truncated = h.truncate(label, 22) if not label_function else label %}
             {% set count = count_label(item['count']) if count_label else ('(%d)' % item['count']) %}
             <li class="{{ nav_item_class or 'nav-item' }}
-                       {% if item.active %}active{% endif %}">
+                       {% if item.active %}active{% endif %}
+                       {% if loop.index > 5 %}facet-{{ name }} show-more-items{% endif %}">
```
KatiRG, line 38: I recommend not hard-coding 5 here

ckanext/ontario_theme/templates/internal/snippets/facet_list.html
```diff
@@ -47,15 +48,14 @@ <h4 class="ontario-h5">
 
       <p class="module-footer">
         {% if h.get_param_int('_%s_limit' % name) %}
-          {% if h.has_more_facets(name, c.search_facets) %}
-            <a href="{{ h.remove_url_param('_%s_limit' % name, replace=0, extras=extras, alternative_url=alternative_url) }}"
-               role="button"
-               class="facets-show-more ontario-button ontario-button--tertiary">{{ _('Show more') }}</a>
+          {% if items|length > 5 %}
+            <button id="more-{{ name }}"
```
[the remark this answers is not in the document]
Alexandra-Green: How do you suggest we do that, accounting for translations?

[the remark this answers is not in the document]
KatiRG: In the JS function that toggles the classes, you can also toggle the button text. If the language is not accessible from the script itself, you can passing in the language as an argument to handle the translation.

KatiRG (approved):

PR #409, opened 2023-09-21
[Task]: Schema Markup

Alexandra-Green:
## What this PR accomplishes

- Adds schema markup to the dataset page and resource page
- Adds the HTML meta tags to the head

## Issue(s) addressed

- Dataset tags, that are part of the dataset metadata, are now part of to the html meta tags for both dataset & resource pages.
- All dataset and resource pages have html meta tags

## What needs review

- On Test, use the [schema.org validator](https://validator.schema.org/) and [Rich Text Results](https://search.google.com/test/rich-results) to make sure there are no errors in the markup
  * There is one known error on the dataset page, it is for the creator itemprop.
- Review code and additional markup placement


ckanext/ontario_theme/templates/internal/package/snippets/resources_list.html
```diff
@@ -108,4 +108,7 @@ <h2>{{ _("Supporting Files") }}</h2>
       {% endif %}
     {% endif %}
   {% endblock resource_list %}
+  <meta itemprop="dateModified" content="{{ pkg.metadata_modified }}"/>
```
aimalkhan, line 111: do we need `dateModified` in both locations (here and resource_item.html:51)?

aimalkhan (approved):
👍 looks great!

PR #415, opened 2023-09-28
[Task]: AI and COVID-19 group pages minor markup fixes

Alexandra-Green:
## What this PR accomplishes

- Allows HTML tags in the groups markdown tag. This is to only allow for abbreviations to be added using `<abbr>` tags
- Removes the `<h1>` in the body, as there is already an `<h1>` on the page.

## Issue(s) addressed

- Allowing the HTML tags will allow me to go in and add the necessary abbreviation tags to the AI group description. Once that is done, the ticket will be ready for QA by designers

## What needs review

- There is only 1 `<h1>` tag on the groups about page. 
- Adding inline HTML tags in the groups description works.


KatiRG:
- [ ] There is only 1 <h1> tag on the groups about page.
- [ ] Adding inline HTML tags in the groups description works

KatiRG (approved):

PR #419, opened 2023-10-04
[Task]: Add missing nav role and aria labels to current page

Alexandra-Green:
## What this PR accomplishes

- Adds navigation role to header of medium and desktop screens
- Adds aria-current="page" state to the current active page in all screens
- Adds new aria-labels for header as seen in Design System's latest release
- Fixes translation for Menu and Close

## Issue(s) addressed

- header missing role=”navigation”, so active page does not have active status in screen reader

## What needs review

- The `<a>` tag of the active page has `aria-current="page"` on all screens
- Aria labels of header match aria labels of Design System application header 
  - aria-label='Show navigation menu' and aria-label="Hide navigation menu"
- Header has role="navigation" in desktop and medium screens and mobile screens have the `<nav>` tag


KatiRG:
- [ ]     The a tag of the active page has aria-current="page" on all screens
- [ ]     Aria labels of header match aria labels of Design System application header: aria-label='Show navigation menu' and aria-label="Hide navigation menu"
- [ ]     Header has role="navigation" in desktop and medium screens and mobile screens have the nav tag

Alexandra-Green:
@KatiRG I've decided to remove the nav role and give the `<section>` an accessible name that alerts users they're entering a region with navigation elements. I don't know what the best practice is, but it removes the warning that our content is outside of landmark roles. We may need to get advice from the ACOE on it.

ckanext/ontario_theme/templates/internal/header.html
```diff
@@ -204,7 +205,7 @@
                 <button class="ontario-hide-for-large ontario-header__menu-toggler ontario-header-button ontario-header-button--with-outline {{ 'fr' if current_lang == 'fr' }}"
                         id="ontario-header-menu-toggler"
```
KatiRG, line 206: There are 2 buttons active at the same time with the same id's:
- id="ontario-header-menu-toggler"
- id="ontario-header-nav-toggler"

Screenshot 2023-10-06 at 09 58 08
Screenshot 2023-10-06 at 09 58 30

when on mobile screen for the same "X" button:
Screenshot 2023-10-06 at 09 59 20

Alexandra-Green: This is consistent with the design system [application header](https://designsystem.ontario.ca/components/preview/application-header.html)

KatiRG: huh

KatiRG (approved):

PR #421, opened 2023-10-04
[Task]: group page minor fixes

Alexandra-Green:
## What this PR accomplishes

- Removes table headers for screen reader only
- Adds "dataset" text to number of datasets per group for screen readers only

## Issue(s) addressed

- DATA-1289

## What needs review

- `aria-hidden="true"` attributes on "Groups" and "Datasets" table headers
- Screen reader only text of "dataset" or "datasets" for each number of datasets for each group


KatiRG:
- [ ] aria-hidden="true" attributes on "Groups" and "Datasets" table headers
- [ ] Screen reader only text of "dataset" or "datasets" ("jeu de données" / "jeux de données") for each number of datasets for each group

KatiRG (approved):

PR #422, opened 2023-10-05
[Task]: Add description lists to About page content 

Alexandra-Green:
## What this PR accomplishes

- Adds descriptions lists to French and English File types and Metadata sections

## Issue(s) addressed

- File types and metadata on [About page](https://data.ontario.ca/about) are lengthy and don’t provide reader with information about what sections accomplish

## What needs review

- In both English and French, "File types" and "Metadata" list of terms are encased in definition lists. Each term should be in a `<dt>` tag and the associated paragraphs should be wrapped in `<dd>` tags.
- There should be 3 definition lists on each About page (EN & FR).


aimalkhan (approved):
👍

PR #426, opened 2023-10-13
[Task]: Adds 'datasets' to topic count on homepage

Alexandra-Green:

- Add aria label to numbers with “datasets” to be read aloud (for example, “Finance (39)” should be read by screen reader as “Finance 39 datasets

## What needs review

- Screen readers read out 'datasets' (and 'jeux de donnees' on French site) along with the topic counts on the homepage.


KatiRG:
Screen reader reads correctly in French: "112 jeux de données"

but in English reads "112 list item"

KatiRG (approved):

PR #434, opened 2023-10-27
[Task]: Remove the RSS box from the Dataset page

Alexandra-Green:
## What this PR accomplishes

- Removes RSS box from the Dataset page

## Issue(s) addressed

-RSS function has an extremely low engagement rate and is an unfamiliar functionality to the majority of our users. Removing the “Keep updated“ box from the right side bar will allow for more important content to take it’s place

## What needs review

- User doesn’t see the “Keep updated“ box on the Dataset page


aimalkhan (approved):
👍

PR #436, opened 2023-11-01
[Task]: Add aria labels to pagination

Alexandra-Green:
## What this PR accomplishes

- Adds `page,py` to override `Page` class in CKAN core's `pagination.py`
- Adds necessary labels to each pagination link
- Changes next and previous page symbol to DS single chevron 
- Removes link tag from `...` list items
- Adds `aria-current` role to current page.
- Translates all aria-labels
- adds navigation role to pagination wrapper
- Updates `dominate` package

## Issue(s) addressed

- DATA-1299

## What needs review

- Pagination has role navigation
- aria-current = "page" for current page
- aria-label for next and previous pages
- next and previous pages use single chevron
- overall accessibility on pagination is good
- design is responsive

**Note:** you'll have to install `requirements.txt` to get the some aria labels working correctly


requirements.txt
KatiRG, line 1: what is dominate?

KatiRG:
Screen reader is saying "One link list item", "Two link list item", ... "right-pointing double-angle quotation mark"

requirements.txt
Alexandra-Green: Python package CKAN core uses to create the HTML tags in their `pagination.py`

aimalkhan: Should it not be already installed since it is being used by ckan-core?

Alexandra-Green: Yes, it is already installed in ckan core. It is on an older version. It needs to at least be on 2.6.0 which is the version specified in the requirements.txt

KatiRG (approved):

aimalkhan (approved):
👍

PR #437, opened 2023-11-03
[Task]: Groups page should be aligned to the Ministries page

Alexandra-Green:
## What this PR accomplishes

- Matches Ministries list items design to the table in the Groups page
- Switches Groups page design from div to list
- Makes entire list item row clickable as link
- Gives accessible name and description (if needed) to link 

## Issue(s) addressed

- When a user places their mouse cursor over the row - the whole row changes the look to mouse-hover and acts as a hyperlink.
- Mirrors the design from the Groups page (a table) on the Ministries page (a list)

## What needs review

- The design of the groups list on the Groups page and Ministries page look identical with the exception of Groups page having a short group description underneath the group name. 
- The interaction of the groups table: on mouse hover on both pages whole row is getting highlighted and acts as a hyperlink.

### DIFF
--- a/ckanext/ontario_theme/fanstatic/common.css
+++ b/ckanext/ontario_theme/fanstatic/common.css
@@ -771,6 +771,10 @@ a:focus {
   line-height: 45px;
 }
 
+a.ministry-heading {
+  text-decoration: none;
+}
+
 .ministry-item.media-item:first-child {
   border-top: 1px solid #dddddd;
 }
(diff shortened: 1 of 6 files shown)


KatiRG:
- [ ]      When a user places their mouse cursor over the row - the whole row changes the look to mouse-hover and acts as a hyperlink.
- [ ]      Mirrors the design from the Groups page (a table) on the Ministries page (a list)

Alexandra-Green:
@KatiRG that bottom line is associated with the pagination and not the table, so it will likely be out of scope for this pr.

KatiRG (approved):

PR #456, opened 2023-12-07
[TASK]: Only display "Report an error" on Open datasets

KobikahChandran:
### What this PR accomplishes
"Report an error with this data" banner only appears when the dataset is Open

### Issue(s) addressed
DATA-1348

### What needs review
If the dataset is Open, the banner is at the bottom of the page.
If the dataset is either Under Review or Restricted, the banner does not appear.


aimalkhan:
@Alexandra-Green is this PR dealing with just 1348? seems like there are commits from two different people in there.

Alexandra-Green:
@aimalkhan it's because it's going into main instead of develop. Can we change the default for merges to go into develop?

ckanext/ontario_theme/templates/internal/package/read.html
```diff
@@ -84,7 +84,9 @@ <h1 itemprop="name">
           {% snippet "package/snippets/tags.html", tags=pkg.tags %}
         {% endblock package_tags %}
 
-        {% snippet "package/snippets/report_an_error.html" %}
+        {% if pkg.resources %}
```
Alexandra-Green, line 87: If a dataset is not open but has a resource, the banner still shows up. Changing this line to something like `if pkg.resources and the package's "access_level" == open` will prevent the banner from showing up

Alexandra-Green (changes requested):

ckanext/ontario_theme/templates/internal/package/read.html
```diff
@@ -84,7 +84,9 @@ <h1 itemprop="name">
           {% snippet "package/snippets/tags.html", tags=pkg.tags %}
         {% endblock package_tags %}
 
-        {% snippet "package/snippets/report_an_error.html" %}
+        {% if pkg.resources ['access_level'] == 'open' %}
```
Alexandra-Green, line 87:
```suggestion
        {% if pkg.resources and pkg['access_level'] == 'open' %}
```

Alexandra-Green (approved):
Looks good!

PR #458, opened 2023-12-11
[Task]: Update <title> tags

Alexandra-Green:
## What this PR accomplishes

- Translates `<title>` tag according to language selected

## Issue(s) addressed

- DATA-1374 as noted in DATA-1349

## What needs review

- The `<title>` tags across website have the correct translation of the title in the html and the titles in the browser tab are correct according to the page's language.


KatiRG:
- [ ] The <title> tags across website have the correct translation of the title in the html and the titles in the browser tab are correct according to the page's language.

KatiRG (changes requested):
There are a couple of other files using `g.site_title`:

- https://github.com/ongov/ckanext-ontario_theme/blob/DATA-1374_site_title/ckanext/ontario_theme/templates/external/header.html#L7
- https://github.com/ongov/ckanext-ontario_theme/blob/DATA-1374_site_title/ckanext/ontario_theme/templates/internal/user/edit_user_form.html#L17

Alexandra-Green:
> There are a couple of other files using `g.site_title`:
> 
> * https://github.com/ongov/ckanext-ontario_theme/blob/DATA-1374_site_title/ckanext/ontario_theme/templates/external/header.html#L7

I don't the first part of the `if.. else..` is being used, we don't have a logo, so it seems redundant. This would file would qualify for some clean up

> * https://github.com/ongov/ckanext-ontario_theme/blob/DATA-1374_site_title/ckanext/ontario_theme/templates/internal/user/edit_user_form.html#L17

This one refers to an email subscription, which I'm not even sure how to find, because the option is not available in the user edit form. This one also doesn't seem to ever be used

KatiRG:
> > There are a couple of other files using `g.site_title`:
> > 
> > * https://github.com/ongov/ckanext-ontario_theme/blob/DATA-1374_site_title/ckanext/ontario_theme/templates/external/header.html#L7
> 
> I don't the first part of the `if.. else..` is being used, we don't have a logo, so it seems redundant. This would file would qualify for some clean up
> 
> > * https://github.com/ongov/ckanext-ontario_theme/blob/DATA-1374_site_title/ckanext/ontario_theme/templates/internal/user/edit_user_form.html#L17
> 
> This one refers to an email subscription, which I'm not even sure how to find, because the option is not available in the user edit form. This one also doesn't seem to ever be used

I vote for file cleanup. Test to see if they can be removed without breaking anything. The original files are part of core CKAN so those parts can always be retrieved if needed in future.

KatiRG (approved):

PR #463, opened 2023-12-28
[Task]: Update the tab labels on ministry and dataset pages 

KobikahChandran:
### What this PR accomplishes
Ministry and dataset labels have been updated so that "Dataset(s)" labels are now "Asset(s)" and "Activity Stream" is now "History"

### Issue(s) addressed
DATA-1335

### What needs review
- Different ministry pages and individual datasets should be checked to see if the changes consistently appear
- Tab labels match page labels
- Spelling and French labels appear


Alexandra-Green (changes requested):
You'll need to add the changes to the Groups tab when on the "History" in page tab, the tab (top left) still says "Activity Stream" 
image

ckanext/ontario_theme/templates/internal/group/activity_stream.html
```diff
@@ -0,0 +1,12 @@
+{% extends "group/read_base.html" %}
+
+{% block subtitle %}{{ _('History') }} {{ g.template_title_delimiter }} {{ super() }}{% endblock subtitle %}
+
+{% block primary_content_inner %}
+  <h1 class="hide-heading">
+    {% block page_heading -%}
+      {{ _('History') }}
+    {%- endblock %}
```
Alexandra-Green, line 9: Nitpick: Can we change this to `endblock page_heading` ?

Alexandra-Green (changes requested):
You'll need to add the changes to the Groups tab when on the activity stream

Alexandra-Green (approved):

PR #464, opened 2024-01-02
[Bug]: French sorting for datasets does not lead to correct results

Alexandra-Green:
## What this PR accomplishes

- Uses solr to sort French package/dataset titles on search pages in correct order, including accents and lowercase letters
- Improves performance of French name ascending/descending sort on search pages

## Issue(s) addressed

- DATA-1162

## What needs review

- Consult ticket for directions on how to implement changes
- Check that the datasets on all search pages are sorted correctly in French when `Name Ascending` and `Name Descending` are chosen.


aimalkhan (approved):
👍

PR #473, opened 2024-01-09
[Task]: Add a ckan validator to trim whitespace from EN and FR titles

Alexandra-Green:
## What this PR accomplishes

- Creates validator to strip whitespace from fluent fields.
- Adds validator to dataset and resource schemas for titles and names

## Issue(s) addressed

- DATA-1392

## What needs review

- When creating/updating a dataset and/or resources with EN and FR titles that start or end with a space, the whitespaces get trimmed.
- When sorting `Name Ascending`, these datasets don't appear out of sort order at the beginning of the list.


KatiRG:
- [ ]     When creating/updating a dataset and/or resources with EN and FR titles that start or end with a space, the whitespaces get trimmed.
- [ ]     When sorting Name Ascending, these datasets don't appear out of sort order at the beginning of the list.

KatiRG (approved):

PR #516, opened 2024-02-22
[Task]: Search accuracy - Case insensitivity

Alexandra-Green:
## What this PR accomplishes

- Tags were of type String in solr, this PR changes tags to be of type "text_case_insensitive", a custom Text field.
- String fields are stored exactly as they are and cannot be modified while Text fields can be modified and can go through lower-casing.

## Issue(s) addressed

- DATA-871

## What needs review

- Searching terms like "Health" and "health" produce the same results in the same order.
- Requires solr reconfiguring


aimalkhan:
@Alexandra-Green can you please bump up the version in `setup.py` to `2.1.19` - the current pipeline configuration has caused an issue that it will resolve.

aimalkhan (approved):
looks good! 👍

PR #523, opened 2024-03-07
[Bug]: Right side boxes are displayed incorrectly on large screens

Alexandra-Green:
## What this PR accomplishes

- Removes ckan grid system and replaces is with DS grid system on dataset and resource page
- Fixes right hand side boxes error on large screens between 1170px-1199px wide

## Issue(s) addressed

- DATA-1435

## What needs review

- On screens 1170px-1199px wide, the grey boxes are on the right side of the main content of the dataset and resource pages.
- All content on both types of pages are responsive.


KobikahChandran (commented):
Looks perfect!

PR #524, opened 2024-03-07
[Bug]: Additional information table overlaps side boxes

Alexandra-Green:
## What this PR accomplishes

- Adds `word-wrap: break-all` to the source url in the additional table
- * This is a known issue with html tables and urls, works fine on Firefox

## Issue(s) addressed

- DATA-1434

## What needs review

- On google chrome/edge/safari, the additional information table does not overlap the side boxes when there is a long URL. 
- * example url: 
    https://www.health.gov.on.ca/en/pro/programs/emergency_health/land/responsetime.aspx

### DIFF
--- a/ckanext/ontario_theme/fanstatic/common.css
+++ b/ckanext/ontario_theme/fanstatic/common.css
@@ -851,6 +851,7 @@ a:focus {
 }
 
 @media screen and (max-width: 32.5em) {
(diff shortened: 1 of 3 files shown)


KobikahChandran (commented):
Looks great! (was only able to check on Chromium)

PR #542, opened 2024-03-21
[Task]: Add push to datastore functionality to button on Data Dictionary page

Alexandra-Green:
## What this PR accomplishes

- Pushes to datastore when a data dictionary is successfully validated

## Issue(s) addressed

- DATA-1451

## What needs review

- Datastore is updated and file hash is not ignored when the data dictionary is successfully validated. 
- If you refresh the page, the column type in parentheses will have updated to the overrode type
- If you go to the upload to datastore page, the output has `Fields do not match; new type for <column field> field is <new type> but existing type is <existing type>`


KatiRG (approved):
Looks good. I tested for small files, and large (16MB & 32MB) files.

PR #543, opened 2024-03-22
[Task]: Delete Help page

KobikahChandran:
## What this PR accomplishes
- Deletes Help page
- Deletes all references to the Help page from the About page

## Issue(s) addressed
- DATA-1341

## What needs review
- Help page is removed completely including from navigation
- All references to the Help page are removed
- Help page goes to 404 page and not 500


Alexandra-Green (changes requested):
Can you remove the Colby help page and it's translations from the .po & .pot files as well?

ckanext/ontario_theme/plugin.py
```diff
@@ -181,12 +181,6 @@ def resource_display_name(resource_dict):
 
 ckan.lib.helpers.resource_display_name = resource_display_name
 
-def help():
-    '''New help page for site.
-    '''
-    return render_template('home/help.html')
-
-
 def csv_dump():
```
Alexandra-Green, line 184: Can you add an additional new line above line 184? So there are 2 new lines before `def csv_dump():`

Alexandra-Green:
@aimalkhan do we need to update the version # in the `setup.py`?

aimalkhan:
> @aimalkhan do we need to update the version # in the `setup.py`?

@KobikahChandran update it to 2.1.32 please.

Alexandra-Green (approved):

PR #545, opened 2024-04-03
Data Validation MVP 1: placeholder page for Step 4 publish

KatiRG:
## What this PR accomplishes
Connects the "Run validator to continue" button in the Data Dictionary page to a new page which will contain a review of the metadata and data dictionary fields. At the moment, this page is just a placeholder for the future content. 

This placeholder page will only be rendered when the data dictionary validation completes successfully. If the data dictionary validation fails, the page stays on the Data Dictionary page and the validation error report is shown, as before.

## What needs review
- Test that the workflow from the Data Dictionary page works as expected for 2 cases:
    - a csv file with valid data types
    - a csv file with a data type error (e.g. string in integer column)
 - Make sure that `pkg` variable is passed to the placeholder page (the content of `pkg` should display on the page)

### DIFF
--- a/ckanext/ontario_theme/plugin.py
+++ b/ckanext/ontario_theme/plugin.py
@@ -937,6 +937,11 @@ ckanext.ontario_theme:schemas/ontario_theme_organization.json
 ckanext.ontario_theme:schemas/ontario_theme_group.json
 """
 
+def new_resource_publish(id, resource_id):
+    '''New page for submitting new resource for publication.
+    '''
+    pkg_dict = toolkit.get_action(u'package_show')(None, {u'id': id})
+    return render_template('/package/new_resource_publish.html', id=id, resource_id=resource_id, pkg_dict=pkg_dict)
 
 class OntarioThemePlugin(plugins.SingletonPlugin, DefaultTranslation):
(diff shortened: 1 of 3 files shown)


ckanext/ontario_theme/plugin.py
```diff
@@ -937,6 +937,11 @@ def update_config(self, config_):
 ckanext.ontario_theme:schemas/ontario_theme_group.json
 """
 
+def new_resource_publish(id, resource_id):
+    '''New page for submitting new resource for publication.
+    '''
+    pkg_dict = toolkit.get_action(u'package_show')(None, {u'id': id})
+    return render_template('/package/new_resource_publish.html', id=id, resource_id=resource_id, pkg_dict=pkg_dict)
```
aleeexgreeen, line 944: This is in between 2 classes, can we move it above the `OntarioThemeExternalPlugin` class, so the function is with all the other standalone functions?

KatiRG: Sure. I'll move it under `def help()` since it is also for a new page

aleeexgreeen (approved):

PR #546, opened 2024-04-08
[Task]: Add Step Indicators to workflow pages

aleeexgreeen:
## What this PR accomplishes

- Adds step indicator to new resource, data dictionary, and publishing pages

## Issue(s) addressed

- DATA-1445

## What needs review

- The back button redirect users to the logical previous step
- Step indicator is present on aforementioned workflow pages

### DIFF
--- a/ckanext/ontario_theme/templates/internal/package/new_resource_not_draft.html
+++ b/ckanext/ontario_theme/templates/internal/package/new_resource_not_draft.html
@@ -11,14 +11,7 @@
 {% endblock secondary %}
 
 {% block pre_primary %}
-  <div class="ontario-step-indicator">
-    <div class="ontario-columns ontario-small-12">
-      <div class="ontario-step-indicator--without-back-button">
-        <span class="ontario-h4">Step&nbsp;1 of&nbsp;3</span>
-      </div>
-      <hr />
-    </div>
-  </div>
+  {% snippet "package/snippets/step_indicator.html", step=1 %}
 {% endblock pre_primary %}
 
 {% block form_title %}
--- a/ckanext/ontario_theme/templates/internal/package/new_resource_not_draft.html
+++ b/ckanext/ontario_theme/templates/internal/package/new_resource_not_draft.html
@@ -7,11 +7,27 @@
   </li>
 {% endblock breadcrumb_content %}
 
-{% block secondary_content %}
-  {% snippet "package/snippets/disclaimer.html" %}
-  {{ super() }}
-{% endblock secondary_content %}
+{% block secondary %}
+{% endblock secondary %}
+
+{% block pre_primary %}
+  <div class="ontario-step-indicator">
+    <div class="ontario-columns ontario-small-12">
+      <div class="ontario-step-indicator--without-back-button">
+        <span class="ontario-h4">Step&nbsp;1 of&nbsp;3</span>
+      </div>
+      <hr />
+    </div>
+  </div>
+{% endblock pre_primary %}
+
+{% block form_title %}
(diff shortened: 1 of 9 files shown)


KatiRG:
Clicking the "back" button from the Data Dictionary page
Screenshot 2024-04-09 at 11 29 31

directs to the Edit resource page:
Screenshot 2024-04-09 at 11 30 36

instead of the Step 1 of 3 page:
Screenshot 2024-04-09 at 11 31 19

ckanext/ontario_theme/templates/internal/package/new_resource_publish.html
```diff
@@ -1,16 +1,20 @@
-{% extends "package/base.html" %}
+{% extends "package/resource_edit_base.html" %}
```
KatiRG, line 1: Can you revert back to `package/base.html`? This will be necessary for the actual content of this step.

ckanext/ontario_theme/templates/internal/package/resource_edit_base.html
KatiRG, line 1: Copy the code added to `resource_edit_base.html` and paste in `base.html`.

aleeexgreeen: `base.html` is used on datasets and resources and `resource_edit_base.html` is used only on resources, we are not touching the dataset admin page yet, right?

ckanext/ontario_theme/templates/internal/package/new_resource_publish.html
aleeexgreeen: Since `resource_edit_base.html` extends `base.html` this change won't break anything, but it will keep the page consistent with all the resource edit pages (i.e. the pages that use the step indicator and any pages from the tabs in the edit resource page)

aleeexgreeen:
@KatiRG Step 1 is create a new resource. If the user is sent back to step 1, they will have to create a new resource rather than edit the one they are trying to publish.

ckanext/ontario_theme/templates/internal/package/new_resource_publish.html
aleeexgreeen: I changed the name from `res` to `resource` in the `plugin.py` (and set res to equal resource in the template to avoid changing all instances of res) and it was fine.

ckanext/ontario_theme/templates/internal/package/resource_edit_base.html
KatiRG: No not yet

KatiRG:
> @KatiRG Step 1 is create a new resource. If the user is sent back to step 1, they will have to create a new resource rather than edit the one they are trying to publish.

True. We have not redesigned the Edit workflow to match the updated workflow, so we'll leave this as is for this ticket.

KatiRG:
- [x]     The back button redirect users to the logical previous step
- [x]     Step indicator is present on aforementioned workflow pages

ckanext/ontario_theme/templates/internal/package/new_resource_not_draft.html
KatiRG, line 1: Where does this file get used?

aleeexgreeen: On the new resource page

KatiRG: The new resource page functions without it. But it's good practice to make all pages consistent for whenever they do get used.

KatiRG (approved):

PR #547, opened 2024-04-09
[Task]: Search bar message should be "## datasets found with the search term "##""

aleeexgreeen:
## What this PR accomplishes

- In the grey search bar display the total number of datasets found as "## datasets found with the search term "##""
- Similarly, for ministries and groups

## Issue(s) addressed

- DATA-1403

## What needs review

- In the grey search bar display the translation for result text "## datasets found with the search term "##"" is correct
- Similarly, for ministries and groups


KatiRG:
- [ ]     In the grey search bar display the translation for result text "## datasets found with the search term "##"" is correct
- [ ]     Similarly, for ministries and groups

KatiRG (approved):

PR #552, opened 2024-04-19
Create and integrate Step 2 page in New Resource workflow

KatiRG:
## What this PR accomplishes
- creates an intermediate route (Step 2) in between Create New Resource page and Data Dictionary page
- new route redirects to the validation report of ckanext-validation which is displayed on Step 2 page
- step indicators in all pages of new workflow have the correct number of total steps (4)
- removes "Save & Add another" button from Step 1 page

## What needs review
Verify that:
- the Step 2 page comes after Step 1 in new workflow when creating a New Resource and shows the success validation message
- step indicators in all pages of new workflow have the correct number of total steps (4)


ckanext/ontario_theme/fanstatic/common.css
aleeexgreeen, line 1: https://github.com/ongov/ckanext-ontario_theme?tab=readme-ov-file#installing-dev-tools-and-pre-commit-hooks-for-development can you follow these instructions, and then save the css file to make sure the css is formatted correctly?

aleeexgreeen (changes requested):

ckanext/ontario_theme/templates/internal/datastore/dictionary.html
```diff
@@ -8,7 +8,7 @@
 
 {% block pre_primary %}
   {% set back_url = h.url_for(pkg.type ~ '_resource.edit', id=pkg.name, resource_id=res.id) %}
```
aleeexgreeen, line 10: Can we change this `back_url` to the correct step 2 (successful validation page)?

KatiRG: We should not be directing back to Step 2, which is an intermediate page, the user should be directed back to Step 1.

ckanext/ontario_theme/resource.py
```diff
@@ -0,0 +1,190 @@
...
+from flask.views import MethodView
+import ckan.model as model
+import six
+import ckan.lib.datapreview as lib_datapreview
+import ckan.plugins as plugins
+import cgi
+
+from ckan.common import _, g, request
+from ckan.views.dataset import (
+    _get_pkg_template, _get_package_type, _setup_template_variables
```
aleeexgreeen, line 15: `_get_package_type, _setup_template_variables` are both imported but never used, we can remove them

aleeexgreeen (changes requested):

aleeexgreeen (approved):

PR #555, opened 2024-04-23
Modify style and position of Continue button in Step 1

KatiRG:
## What this PR accomplishes
Styles the button on Step 1 of the new workflow for resource creation.

## What needs review
Confirm that the button on Step 1 of the new workflow for creating a resource:
- is positioned on the left
- has no padding issues in inactive, hover or click state
- has the DS stylings
- text is "Continue"


ckanext/ontario_theme/fanstatic/common.css
```diff
@@ -3492,4 +3492,8 @@ label::after {
     align-items: center;
     gap: 20px;
   }
+}
+
+.form-actions {
+  overflow: unset;
 }
```
aleeexgreeen, line 3499: You can add `text-align: unset;` here and achieve the same result as adding the class `.pull-left` and remove the class from the button. What do you think?

aleeexgreeen (approved):

PR #560, opened 2024-04-29
rm breadcrumbs from workflow

KatiRG:
## What this PR accomplishes
Removes breadcrumbs from the workflow.

## What needs review
Verify that there are no longer any breadcrumbs in the workflow pages:
- Step 1-4
- when clicking Back button while in any of Step 1-4
- Step 1 of editing an existing resource


aleeexgreeen (changes requested):
image
Breadcrumbs aren't actually removed, you'll need to change the empty blocks to `block breadcrumb` instead of `block breadcrumb_content`

aleeexgreeen (approved):

PR #561, opened 2024-04-29
[Task]: Add Aria labels to list items on dataset pages

aleeexgreeen:
## What this PR accomplishes

- Adds accessible name to resource-item-div link on dataset pages
- Removes additional links in resource-item-div, leading to resource page

## Issue(s) addressed

- DATA-1479

## What needs review

- Resource item div on dataset page has an aria-label
- For non-admin/non-logged in users, the resource-item-div should have only one link (the anchor-div)
- For admin/logged in users, the resource-item-div will have the title, explore button, and edit button as links.


KobikahChandran (approved):
No concerns!

PR #562, opened 2024-04-30
[Task]: Bypass validation and Data Dictionary form for non-CSV resources

aleeexgreeen:
## What this PR accomplishes

- When a non-CSV file is added as a new resource, redirects to Step 2.
- Added new file and blueprint to show validation/upload status as needed.
- Added snippet for successful status page alert.
- Step 2 redirects to data dictionary page with non-CSV files, reworked the data dictionary prepare function

## Issue(s) addressed

- DATA-1444

## What needs review

- Going through workflow with non-csv files work as expected. (no errors, user goes through all steps and content follow figma)
- Going through workflow with csv files work as expected. (no changes)

### DIFF
--- a/ckanext/ontario_theme/templates/internal/package/resource_validation.html
+++ b/ckanext/ontario_theme/templates/internal/package/resource_validation.html
@@ -1,0 +1,41 @@
+{% extends "package/resource_edit_base.html" %}
+
+{%- block subtitle -%}
+    {{ h.resource_display_name(resource) }}
+{%- endblock subtitle -%}
+
+{% block pre_primary %}
+  {% set back_url = h.url_for(pkg_dict.type ~ '_resource.edit', id=pkg_dict.id, resource_id=resource.id) %}
+  {% snippet "package/snippets/step_indicator.html", step=2, back_url=back_url %}
+{% endblock pre_primary %}
+
+{% block form_title %}
+  {{ h.resource_display_name(resource) }}
+  {% if validation.report %}{{ h.get_validation_badge(resource)|safe }}{% endif %}
+{% endblock form_title %}
+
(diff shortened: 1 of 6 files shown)


KatiRG:
- [ ] Going through workflow with csv files work as expected. (no changes)

KatiRG:
x Going through workflow with non-csv files work as expected.

The workflow works for creating a non-CSV resource, but clicking Edit on an existing non-CSV resource results in an internal server error: `Resource "xxx" was not found.`

aleeexgreeen:
> The workflow works for creating a non-CSV resource, but clicking Edit on an existing non-CSV resource results in an internal server error: `Resource "xxx" was not found.`

@KatiRG Try updating a non-csv file on `ckanext_validation_develop` the same error appears. It is due to the update resource button running synchronous validation on `resource_update`.

KatiRG:
@aleeexgreeen There should be no validation running on the update resource button either. Do you want a separate ticket made for that case?

aleeexgreeen:
@KatiRG That should probably be done during DATA-1472

ckanext/ontario_theme/templates/internal/package/resource_validation.html
KatiRG, line 1: Previously, step 2 was redirected to [validation_read.html](https://github.com/ongov/ckanext-validation/blob/develop/ckanext/validation/templates/validation/validation_read.html) in ckanext-validation. Since `resource_validation.html` is replacing `validation_read.html`, can we revert `validation_read.html` to the original?

KatiRG:
Looks good. Pending only the content update.

KatiRG (approved):

PR #571, opened 2024-05-09
Direct Edit resource page into Step 2 of new workflow

KatiRG:
## What this PR accomplishes
The button on the Edit Resource page should direct the user into Step 2 of the new workflow and not into the Dataset page.

## What needs review
- Upload a new CSV resource (Step 1), then test all the Edit pathways to ensure that the workflow proceeds from the Edit page into Step 2 as expected, ie:
    - from Steps 2, 3, and 4, click the Back link in the step indicator to get to the Edit page, make an edit, and proceed
    - after publishing a new resource, click the Edit button on the Dataset page, make an edit, and proceed

- The button text on the Edit page should be "Continue" and should have Design System styles. Note that in this PR, the button is still positioned at the bottom right of page. Another ticket will address the position of the buttons.

## What should not be reviewed
Do not test the Edit page for a non-CSV resource. This case will be handled in another PR in our ckanext-validation fork.


ckanext/ontario_theme/templates/internal/package/snippets/resource_edit_form.html
```diff
@@ -0,0 +1,14 @@
...
+{% set include_metadata = include_metadata or false %}
+
+{% extends resource_form_snippet %}
+
+{% block stages %}{% endblock %}
+
+{% block previous_button %}{% endblock %}
+
+{% block again_button %}{% endblock %}
+
```
aleeexgreeen, line 11: You don't need this entire bit of code. Using `{% ckan_extends %}` here and using `{% block save_button %}` works.

ckanext/ontario_theme/templates/internal/package/snippets/resource_edit_form.html
```diff
@@ -0,0 +1,14 @@
...
+{% block save_button %}
+<button class="ontario-button ontario-button--primary" name="save" type="submit">{% block save_button_text %}Continue{% endblock %}</button>
+{% endblock %}
```
aleeexgreeen, line 14: Can you format these blocks? Endblocks should have the block name. Please indent the button and the block and text inside the button and block.

aleeexgreeen (changes requested):

ckanext/ontario_theme/templates/internal/package/snippets/resource_edit_form.html
```diff
@@ -0,0 +1,9 @@
+{% ckan_extends %}
+
+{% block save_button %}
+  <button class="ontario-button ontario-button--primary"
+          name="save"
+          type="submit">
+    {% block save_button_text %}Continue{% endblock %}
```
aleeexgreeen, line 7:
```suggestion
    {% block save_button_text %}
      Continue
    {% endblock save_button_text %}
```

aleeexgreeen (changes requested):

ckanext/ontario_theme/templates/internal/package/snippets/resource_edit_form.html
KatiRG: This is fine, but fyi it is not flagged by `djlint`.

aleeexgreeen: @KatiRG it wouldn't be flagged with an unconfigured djLint. It would be flagged with our pre-commit hooks which is configured for our repo.

aleeexgreeen (approved):

PR #580, opened 2024-05-28
[Bug]: Resources with URL titles overlap with side boxes

aleeexgreeen:
## What this PR accomplishes

- Adds `overflow-wrap: break-word` to the resource page titles
- Adds `word-break: break-all` to the resource page titles with URLs as the titles on mobile
- Replace instance of `word-wrap: break-word` as it is deprecated
- Removed truncation on breadcrumbs on resource page

## Issue(s) addressed

- DATA-1507

## What needs review

- Resource with URL titles do not overflow horizontally on mobile
- On medium and large screens do not overlap with the side boxes
- No truncation on resource page breadcrumbs

### DIFF
--- a/ckanext/ontario_theme/fanstatic/common.css
+++ b/ckanext/ontario_theme/fanstatic/common.css
@@ -1776,6 +1776,11 @@ via submission form. Copied from DS.*/
   float: right;
 }
 
+.resource-main h1 {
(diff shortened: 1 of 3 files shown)


KobikahChandran (approved):
Looks perfect!

PR #581, opened 2024-05-29
[Task]: Make the dropdowns consistent and change the order of Licence & Access level

aleeexgreeen:
## What this PR accomplishes

- Positions Access Level dropdown above the Licence dropdown in dataset admin
- Remove licence options that are not associated with the selected access level

## Issue(s) addressed

- DATA-696 and DATA-1424

## What needs review

- Access level is above the Licence dropdown
- if the license is set to “Not Applicable”, the Access level cannot be “Open”
- if the Access level is “Restricted” or “Under review”, the license must be “Not Applicable”

**_Since "Not Applicable" is not exclusive to either restricted or under review, the option was removed from open (This will satisfy the above condition)_


KatiRG:
- [ ]     Access level is above the Licence dropdown
- [ ]     if the license is set to “Not Applicable”, the Access level cannot be “Open”
- [ ]     if the Access level is “Restricted” or “Under review”, the license must be “Not Applicable”

KatiRG (approved):

PR #585, opened 2024-07-02
Redirect to dictionary page only when table is available in the DataStore

KatiRG:
## What this PR accomplishes
(see DATA-1455)

**UI changes:**
- implements a JS polling function in Step 2 to monitor when `datastore_active` becomes `True` in the `resource_show` API. The poll is set to check every 1000ms. 
- while the function polls, the Design System spinner is shown and the user is informed that the data is being uploaded to database
- when `datastore_active` becomes `True`, the Continue button is displayed in Step 2 and the user can proceed to Step 3 (data dictionary page)

**Backend performance improvement:**
If the user does not specify any data types in Step 3 (data dictionary page), there is no need to send the data file for validation since columns are already stored in the database as text and there is nothing new to check. The code in `datastore.py` was modified to determine if any column type overrides have been specified, and if not, skip directly to Step 4 (publish page).

## What needs review
- test the workflow with small and large CSV files
    - confirm that the process works, that the spinner is displayed as it should, etc
    - try the workflow without overriding the data dictionary to test the optimization (noticeable only for large files)
    - try the workflow with data dictionary overrides to make sure everything works
- test with non-CSV files to make sure nothing has broken


ckanext/ontario_theme/templates/internal/package/resource_validation.html
```diff
@@ -15,6 +15,13 @@
 {% endblock form_title %}
 
 {% block primary_content_inner %}
+  {% set package_id = resource.get('package_id') %}
+  {% set resource_id = resource.get('id') %}
+  {% set action = h.url_for('datastore.dictionary',
+      id=package_id,
+    resource_id=resource.id) %}
+  {% set is_csv = "True" %}
```
aleeexgreeen, line 23: `def resource_validation(id, resource_id):` in the plugin.py, passes the package id as `id` and you can pass the `resource_id` as well. Line 18 and 19 are redundant and you can use the id and resource_id for `set action`

ckanext/ontario_theme/templates/internal/package/resource_validation.html
```diff
@@ -23,19 +30,79 @@
...
+      };
+      poll_datastore();
+    </script>
+
+    <div class="ontario-small-12 ontario-row">
+      <div class="ontario-small-12 ontario-columns">
+        <div id="loadingSpinner"
+             aria-hidden='false'
+             role='alert'
+             aria-live='assertive'>
```
aleeexgreeen, line 84: This line is not needed as `role="alert"` implies aria-live is assertive.

ckanext/ontario_theme/templates/internal/package/resource_validation.html
```diff
@@ -23,19 +30,79 @@
...
+          <div class='ontario-loading-indicator'>
+            <svg id="spinnerSVG"
+                 class='ontario-loading-indicator__spinner-small'
+                 viewBox="25 25 50 50"
+                 xmlns='http://www.w3.org/2000/svg'>
```
aleeexgreeen, line 89: Can all HTML attributes values be in double quotes please?

KatiRG: Will do. I got the code from https://designsystem.ontario.ca/components/detail/loading-indicator.html, which has a mix of single and double quotes. Can you let them know?

ckanext/ontario_theme/templates/internal/package/resource_validation.html
KatiRG: This comes from the DS documentation https://designsystem.ontario.ca/components/detail/loading-indicator.html. Can you raise this with them?

ckanext/ontario_theme/templates/internal/package/resource_validation.html
KatiRG: Actually `resource.id` is already in this page

aleeexgreeen: Good point. In fact, `resource_id` from line 19 is never actually used.

ckanext/ontario_theme/templates/internal/package/resource_validation.html
```diff
@@ -23,19 +30,79 @@
...
   {% else %}
+    {% set is_csv = "False" %}
     {% set success_message = "Data uploaded successfully" %}
     {% snippet "package/snippets/successful_page_alert.html", success_message=success_message %}
+    <a href="{{ action }}">
+      <button id="nextBtn"
+              class="ontario-button ontario-button--primary"
+              name="save"
+              type="submit">Continue</button>
+    </a>
```
aleeexgreeen, line 39: Interactive elements shouldn't be nested within one another for accessibility reasons. You can style the link as a button, add `role=button` and add javascript so the keydown and click events operate as expected or wrap the button in a form with the action being the link.

ckanext/ontario_theme/templates/internal/package/resource_validation.html
```diff
@@ -23,19 +30,79 @@
...
+
+    <a href="{{ action }}">
+      <button id="step2Btn"
+              style="visibility: hidden;
+                     margin-top: 40px;
+                     margin-left: 36px"
+              class="ontario-button ontario-button--primary"
+              name="save"
+              type="submit">Continue</button>
+    </a>
```
aleeexgreeen, line 103: See comment above about nested interactive elements.

ckanext/ontario_theme/templates/internal/package/resource_validation.html
aleeexgreeen, line 1: The difference between the buttons in this file is the ids and style. I'd suggest making the id a variable and placing the contents of the style attribute in `common.css` targeting the button's id, so there is only one instance of the button but how it's displayed depends on the id.

ckanext/ontario_theme/plugin.py
```diff
@@ -46,6 +46,8 @@
 import ckan.lib.uploader as uploader
 import ckan.lib.helpers as h
 
+import ckanext.xloader.interfaces as xloader_interfaces
+
```
aleeexgreeen, line 50: I don't believe this is used anywhere, is it needed?

KatiRG: nopes
left-over from previous experiments

ckanext/ontario_theme/templates/internal/package/resource_validation.html
```diff
@@ -23,19 +28,76 @@
...
+
+    <form "method=post" action="{{ action }}">
+      <button id="step2Btn"
+              style="visibility: hidden;
+                     margin-top: 40px;
+                     margin-left: 36px"
+              class="ontario-button ontario-button--primary"
+              name="save"
+              type="submit">{{ _('Continue') }}</button>
+    </form>
```
aleeexgreeen, line 101: There is a misalignment with the button and the text once the spinner is not visible to users. To align the button and text to the left of the screen, remove the margin-left from the button and the text. Instead of using "visibility: hidden;" use "display: none;" in both the button and the javascript. In the javascript, `visibility: visible` would be `display: block`. You'd remove the `margin-left: 12px` from `.ontario-loading-indicator p.spinner-small` and add `margin-right: 12px` to `.ontario-loading-indicator__spinner-small`

ckanext/ontario_theme/templates/internal/package/resource_validation.html
```diff
@@ -23,19 +28,76 @@
...
+                 xmlns="http://www.w3.org/2000/svg">
+              <circle cx="50" cy="50" r="20" fill="none" stroke-width="4" />
+            </svg>
+            <p class="spinner-small" id="pendingTxt">Uploading data file to database</p>
+          </div>
+        </div>
+      </div>
+    </div>
+
+    <form "method=post" action="{{ action }}">
```
aleeexgreeen, line 93: The way method is defined it is not currently being used and I believe the method should be get in this instance.

ckanext/ontario_theme/templates/internal/package/resource_validation.html
aleeexgreeen: You'll have to move the button form outside the `if is_csv == "True"` statement and I'd recommend moving the success alert snippet out of the `if validation.report` as well.

ckanext/ontario_theme/datastore.py
```diff
@@ -81,57 +81,49 @@ def post(self, id, resource_id):
...
-        # to xloader when status is successful
-        status = get_action(u"resource_validation_show")(
-            {u'ignore_auth': True},
-            {u"resource_id": resource_id}
-        )
-
-        if status['status'] == 'success':
-            get_action("xloader_submit")(
+        has_override = True if [x for x in dict_fields['fields'] if len(x['info']['type_override'])>0] else False
+        if has_override:
```
aleeexgreeen, line 85: There is extra whitespace at the end of this line

ckanext/ontario_theme/datastore.py
```diff
@@ -81,57 +81,49 @@ def post(self, id, resource_id):
...
+                )
+                return h.redirect_to("ontario_theme.new_resource_publish", id=id, resource_id=resource_id)
+            elif not status:
+                return h.redirect_to("ontario_theme.new_resource_publish", id=id, resource_id=resource_id)
+            else:
+                return h.redirect_to("datastore.dictionary", id=id, resource_id=resource_id)
         else:
-            return h.redirect_to("datastore.dictionary", id=id, resource_id=resource_id)
+            return h.redirect_to("ontario_theme.new_resource_publish", id=id, resource_id=resource_id)
+
```
aleeexgreeen, line 129:
```suggestion
```

ckanext/ontario_theme/templates/internal/package/resource_validation.html
KatiRG: This will misalign the text after the spinner disappears:
Screenshot 2024-07-03 at 16 24 10

aleeexgreeen: That's because you are using `visibility: hidden;` for the style of the SVG using `display: none;` will fix the alignment.

ckanext/ontario_theme/templates/internal/package/resource_validation.html
```diff
@@ -23,19 +28,75 @@
...
+              }
+              fixedDelay();
+
+            } else {
+            setTimeout(poll_datastore, 1000);
+            }
+          }
+        }
+      };
+      poll_datastore();
```
aleeexgreeen, line 74: I don't see any particular reason why this needs to be in the HTML rather than it's own js file (ideally this would be a [ckan js module](https://docs.ckan.org/en/2.9/theming/javascript.html#javascript-modules)). We don't typically use HTML in-file javascript. @aimalkhan What do you think, should it be kept in the HTML or moved to a js file in the scripts directory?

KatiRG: The js module are attached to HTML elements in the page. Here we want to call a polling function on page load and not associate with any particular HTML element (until the result of the poll).

aleeexgreeen: Okay, if not a js module, then a regular js file will suffice

KatiRG: It does not seem possible to pass arguments to JS functions from the jinja template other than through an html element. There are some instances of `<script>` tags in core ckan templates.

ckanext/ontario_theme/templates/internal/package/resource_validation.html
KatiRG: We can leave it as `post` as we don't need to send any parameters via the url  
https://developer.mozilla.org/en-US/docs/Learn/Forms/Sending_and_retrieving_form_data

ckanext/ontario_theme/templates/internal/package/resource_validation.html
KatiRG: Done, but the snippet needs to stay inside `if validation.report` since it needs to be above the div containing the indicator and indicator text.

ckanext/ontario_theme/templates/internal/package/resource_validation.html
```diff
@@ -15,27 +15,80 @@
...
 
-  {% set package_id = resource.get('package_id') %}
-  {% set resource_id = resource.get('id') %}
-  {% set action = h.url_for('datastore.dictionary',
-      id=package_id,
-      resource_id=resource.id) %}
- 
-  <form method="get" action="{{ action }}">
-    <button class="ontario-button ontario-button--primary"
+  <form "method=post" action="{{ action }}">
```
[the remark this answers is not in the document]
KatiRG: Why is method=post invalid HTML? Can you please add a reference.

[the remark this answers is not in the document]
KatiRG: Core CKAN is using method=post and not get.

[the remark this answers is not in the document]
aleeexgreeen: `method="post"` is valid `"method=post"` is not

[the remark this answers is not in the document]
KatiRG: "method=post" is your code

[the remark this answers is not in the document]
aleeexgreeen: What do you mean it is my code? In this [commit](https://github.com/ongov/ckanext-ontario_theme/pull/585/commits/4532524ce5b87ba656dfbefc497a414b3e1c927d) the form was changed from a form with `method="get"` to an anchor tag and then in this [commit](https://github.com/ongov/ckanext-ontario_theme/pull/585/commits/8596d02413d969e7b6386afa8d878261c11067cd) it was changed from an anchor tag to a form with `"method=post"`, they were both committed by you.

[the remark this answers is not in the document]
KatiRG: https://github.com/ongov/ckanext-ontario_theme/blob/ckanext_validation_dev_DATA1439_poll/ckanext/ontario_theme/templates/internal/datastore/dictionary.html#L24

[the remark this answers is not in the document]
aleeexgreeen: That would be added as a string which the browser would interpret as `method="post"` screenshot:
image
With `<form "method=post"...` it is invalid to the browser screenshot:
image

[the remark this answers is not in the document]
KatiRG: So your issue is with the quotes and not the method.
It needs to be post because it gets processed in `def post` here: https://github.com/ongov/ckanext-ontario_theme/blob/ckanext_validation_dev_DATA1439_poll/ckanext/ontario_theme/datastore.py#L63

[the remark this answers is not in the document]
aleeexgreeen: The form on the data dictionary page uses POST so that is what is calling the datastore function `def post`. Since the browser was not interpreting the `"method="post""` correctly, the form was defaulting to GET. The data dictionary page worked fine. If you were to try to use `method="post"` you would get an internal server error.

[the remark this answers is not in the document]
KatiRG: Resolved! To recap:
- method=post is not invalid html per se, it was the quotes
- for this button the method should be `get` because it directs to the `def get` method of the dictionary page: https://github.com/ongov/ckanext-ontario_theme/blob/ckanext_validation_develop/ckanext/ontario_theme/datastore.py#L52 (and not the post method, my bad)

aleeexgreeen (approved):
All good !

PR #590, opened 2024-07-31
[Task]: Make mandatory metadata fields

KobikahChandran:
## What this PR accomplishes
- Makes certain metadata fields mandatory when creating a new resource

## Issue(s) addressed
DATA-1519 and DATA-1069

## What needs review
- Resource, Format, Name (EN & FR), Description, Type, Language, Data Last Updated need to be completed when creating a new resource
- Dialogue box indicates which mandatory fields are missing at the top


aleeexgreeen (changes requested):
image
The error text box says URL when a file or link is not provided and does not have an asterisk beside File showing it is a required field. The error indicating the value missing isn't visible until you click file upload or link

PR #593, opened 2024-08-07
bugfix: add extras dict to url

KatiRG:
## What this PR accomplishes
The `h.add_url_param()` function requires a dictionary with 3 elements: `id`, `group_type`, and `is_organization`. This PR adds those parameters in to the empty `extras` object which was causing an error when rendering the Datasets page from the admin panel of the Ministry page.

## What needs review
1. Go to Ministries page and pick a ministry
2. click the Edit button
3. click the Datasets tab. Verify that the page displays.

### DIFF
--- a/ckanext/ontario_theme/templates/internal/organization/bulk_process.html
+++ b/ckanext/ontario_theme/templates/internal/organization/bulk_process.html
@@ -9,7 +9,7 @@
   {% set sorting = [
       (_('Name Ascending'), 'titles asc'),
       (_('Name Descending'), 'titles desc'),
-      (_('Last Modified'), 'data_modified desc'),
+      (_('Last Modified'), 'metadata_modified desc'),
       (_('Relevance'), 'score desc, metadata_modified desc'), ] %}
   {% set extras = {'id': group_dict.id} %}
   {% snippet 'snippets/search_form.html', form_id='organization-datasets-search-form',
--- a/ckanext/ontario_theme/templates/internal/organization/bulk_process.html
+++ b/ckanext/ontario_theme/templates/internal/organization/bulk_process.html
@@ -4,13 +4,14 @@
   {% if request['params'] and 'sort' in request['params'] %}
     {% set sort_by_selected = request.params['sort'] %}
   {% else %}
-    {% set sort_by_selected = "titles asc" %}
+    {% set sort_by_selected = "score desc, metadata_modified desc" %}
   {% endif %}
   {% set sorting = [
       (_('Name Ascending'), 'titles asc'),
       (_('Name Descending'), 'titles desc'),
-      (_('Last Modified'), 'data_modified desc') ] %}
-  {% set extras = {'id': c.page.items[0]['organization']['id']} %}
+      (_('Last Modified'), 'data_modified desc'),
+      (_('Relevance'), 'score desc, metadata_modified desc'), ] %}
+  {% set extras = {'id': group_dict.id} %}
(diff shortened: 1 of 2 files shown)


aleeexgreeen (changes requested):
While this works to remove the error when loading the page, the dropdown does not work correctly (changing the dropdown, sort is added to URL but dropdown remains Name Ascending, consequently, Name Ascending is unable be selected as an option). Because this bug was in relation to the Sort By drop down, we should make sure the dropdown works as well.

To tackle both issues, I recommend creating a `bulk_process.html` in the organizations folder and extending it from ckan.  This will also simplify things because the only variable for the extras dict will be the org id, while `group_type` will always be` 'organization'` and `is_organization` will always be `true`, from this html page. 
Copy the `{% block search_form %}`, add 
```
{% if request['params'] and 'sort' in request['params'] %}
     {% set sort_by_selected = request.params['sort'] %}
{% else %}
     {% set sort_by_selected = "titles asc" %}
 {% endif %}
```
add the extras dictionary to the snippet and lastly, change `title_string` to `titles`

KatiRG:
@aleeexgreeen this works but when landing on the Datasets tab, the list is not actually sorted (and the URL has no sort params) even though the dropdown is defaulted to "Name Ascending":

Screenshot 2024-08-09 at 15 41 24

aleeexgreeen:
> @aleeexgreeen this works but when landing on the Datasets tab, the list is not actually sorted (and the URL has no sort params) even though the dropdown is defaulted to "Name Ascending":
> Screenshot 2024-08-09 at 15 41 24

Good point! I honestly don't know how to fix it, but it is sorting by `score desc, metadata_modified desc` aka relevance, so a quick and easy way would to just add relevance as a sorting option and change the else statement to default to `score desc, metadata_modified desc` 🤣

ckanext/ontario_theme/templates/internal/organization/bulk_process.html
```diff
@@ -0,0 +1,23 @@
...
+  {% if request['params'] and 'sort' in request['params'] %}
+    {% set sort_by_selected = request.params['sort'] %}
+  {% else %}
+    {% set sort_by_selected = "titles asc" %}
+  {% endif %}
+  {% set sorting = [
+      (_('Name Ascending'), 'titles asc'),
+      (_('Name Descending'), 'titles desc'),
+      (_('Last Modified'), 'data_modified desc') ] %}
+  {% set extras = {'id': c.page.items[0]['organization']['id']} %}
```
aleeexgreeen, line 13:
```suggestion
  {% set extras = {'id': group_dict.id} %}
```
Core ckan has this `group_dict.id` for the organization id handy that you can use.

KatiRG:
This is working now, although on my local the "Last modified" does not actually list the datasets in correct order. The parameters are correct in the url: `organization/bulk_process/agriculture-food-and-rural-affairs?q=&sort=data_modified+desc`

Can you check on your side?

aleeexgreeen:
> This is working now, although on my local the "Last modified" does not actually list the datasets in correct order. The parameters are correct in the url: `organization/bulk_process/agriculture-food-and-rural-affairs?q=&sort=data_modified+desc`
> 
> Can you check on your side?

Good catch, it should be `metadata_modified+desc`, you can change last modified to `(_('Last Modified'), 'metadata_modified desc')`

aleeexgreeen (approved):
Looks and works great!

PR #594, opened 2024-08-09
Display same Step 1 title for resource adding or editing

KatiRG:
## What this PR accomplishes

- Changes the title of Step 1 for new resources and for editing existing resources to be the same. The title should be: "Add or edit data resource" for both pages.
- Consolidates the text of the browser tab such that text starts with "Add or edit" instead of "Add resource" (new resources) or "Edit" (existing resources). The rest of the tab text should remain as before.

## What needs review
Confirm that:
- Step 1 page `h1` title is `Add or edit data resource` both when creating a new resource and when editing an existing resource
- Step 1 browser tab reads `Add or edit - {dataset name} - {ckan site name}` for new resources and `Add or edit - {resource name} - {dataset name} - {ckan site name}` for existing resources

### DIFF
--- a/ckanext/ontario_theme/templates/internal/package/new_resource_not_draft.html
+++ b/ckanext/ontario_theme/templates/internal/package/new_resource_not_draft.html
@@ -1,5 +1,9 @@
 {% ckan_extends %}
 
+{% block subtitle %}
+  {{ _('Add or edit') }} {{ g.template_title_delimiter }} {{ h.dataset_display_name(pkg) }}
+{% endblock %}
+
 {% block breadcrumb %}
 {% endblock breadcrumb %}
 
@@ -11,7 +15,7 @@
 {% endblock pre_primary %}
 
 {% block form_title %}
-  Add or edit data resource
+  {{ _('Add or edit data resource') }}
 {% endblock form_title %}
 
 {% block form %}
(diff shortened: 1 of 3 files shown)


aleeexgreeen (commented):
Did you was to change the tab titles as well? The new resources page window tab says "Add resource" and the edit page window tab says "Edit"

aleeexgreeen (changes requested):
This works for the most part, but it's unclear if this is the best approach. It deviates from the typical approach, which is changing the tab text through `block subtitle` in the html file. It also has unintended consequences of affecting the tab text of the ministries and groups edit pages only, but not the creation pages. When creating a dataset and moving to the add data step, the tab text is condensed and doesn't particularly make sense with the page, (it was previously, "Add data to the dataset") see screenshot: image

If we were to use the typical approach, 
- we'd add `{% block subtitle %}{{ _('Add or edit') }} {{ g.template_title_delimiter }} {{ h.dataset_display_name(pkg) }}{% endblock %}` to `new_resource_not_draft.html` 
- we'll have to add the file `resource_edit.html` to our repo add `{% ckan_extends %}
{% block subtitle %}{{ _('Add or edit') }} {{ g.template_title_delimiter }} {{ h.resource_display_name(res) }} {{ g.template_title_delimiter }} {{ h.dataset_display_name(pkg) }}{% endblock %}` as the contents of the file.

Considering we will likely have to change the tab text of all the flow pages, it may be best to go with the typical approach.

aleeexgreeen (approved):
Works perfectly!

PR #596, opened 2024-08-15
[Task]: Redesign Dataset pages for accessibility

aleeexgreeen:
## What this PR accomplishes

Makes dataset page more accessible
- Gives tabs group the role `tablist` so screen readers are aware the links are tabs and which tab they are currently on (repeated for tabs on ministries and groups pages for consistency)
- Adds the dataset H1 to every tab page
- Adds `<aside>` after `<article>` on mobile screens instead of using css' `order`, which does not change order in HTML (causing the aside to be announced first, when it is the last thing on the page visually)

## Issue(s) addressed

- DATA-1537

## What needs review

- Dataset name is the only H1 on each tab page
- Tab group has `role="tablist"` and tabs have `role="tab"`. The active tab should have `aria-selected="true"`
- On mobile, Ministry and licence sidebar is being displayed by an `<aside>` displayed after the `<article>` in the HTML


KatiRG:
> Dataset name is the only H1 on each tab page

Confirmed

KatiRG:
> Tab group has role="tablist" and tabs have role="tab". The active tab should have aria-selected="true"

Confirmed

KatiRG:
> On mobile, Ministry and licence sidebar is being displayed by an <aside> displayed after the <article> in the HTML

confirmed

KatiRG (approved):

PR #598, opened 2024-08-19
[Story]: Add checkboxes to the items in the search filters

aleeexgreeen:
## What this PR accomplishes

- Separates access level from filter sidebar and adds it to the search bar
- Defaults to open upon opening the dataset search page or an individual group / ministry page
- Defaults to open when all checkboxes are unchecked
- Reduce templates rendered on the dataset page (improving marginally the performance of page)
- Use of "search term" instead of "keyword(s)" and the use of "total"
- Clear filters button and filter facets now submit form instead of redirecting to links with filters as params in URL
- Search reset "x" now works when typing in a empty search box
- Adds text to empty search with additional datasets are present in different access levels
- Removes duplicate forms wrapping the `search_form.html`
- Adds callout with search term (if present) and access level(s) selected
- Adds an escape to the query string for search results

## Issue(s) addressed

- DATA-1083, DATA-1272, DATA-1514

## What needs review

- All filters should allow multiple checkboxes to be selected and operate as OR.
- Relationship between filters should operate as AND
- Callout is updated with every checkbox selected. When all checkboxes are selected "All levels" is displayed
- Callout is updated when there is no results. Callout is updated when there is a search term present
- The Access level filter should contain 3 items: Open, Under review and restricted. In this order
Filters sidebar:
- All selected checkboxes should be visible, including when there are more than 5 items (will not be bumping it to the top of list)
- If there are more than 5 items selected than the default view be extended to more than 5 to include all selected items.
Access level Zero results:
- if there are zero results in a search with all 3 checkboxes selected Display message “Please try another search or reach out to us if you think data is missing.“
- if there are zero results in a search for the selected checkbox(es) and there are datasets available in unselected checkboxes, there should be a message with a link to the search results of the access level with datasets displayed in the body of the page
Other things needing review:
- Translations are present and correct
- implementation matches Figma design (to the best of my ability, Figma screens are not consistent)
- The number of datasets on homepage equals total dataset amount and not open dataset amount
- The number of datasets per on each Ministry on the ministry page is the total dataset amount and not open dataset amount
- Navigating the website, there are no internal server errors
- Clicking the search reset "x" when typing in a empty search box, resets the search box to empty

Tests should be done on individual ministry and group pages and the dataset search page


KatiRG:
- [ ]     All filters should allow multiple checkboxes to be selected and operate as OR.
- [ ]     Relationship between filters should operate as AND
- [ ]     Callout is updated with every checkbox selected. When all checkboxes are selected "All levels" is displayed
- [ ]     Callout is updated when there is no results. Callout is updated when there is a search term present
- [ ]     The Access level filter should contain 3 items: Open, Under review and restricted. In this order

Filters sidebar:
- [ ]     All selected checkboxes should be visible, including when there are more than 5 items (will not be bumping it to the top of list)
- [ ]     If there are more than 5 items selected than the default view be extended to more than 5 to include all selected items.

Access level Zero results:
- [ ]     if there are zero results in a search with all 3 checkboxes selected Display message “Please try another search or reach out to us if you think data is missing.“
- [ ]     if there are zero results in a search for the selected checkbox(es) and there are datasets available in unselected checkboxes, there should be a message with a link to the search results of the access level with datasets displayed in the body of the page

Other things needing review:
- [ ]     Translations are present and correct
- [ ]     implementation matches Figma design (to the best of my ability, Figma screens are not consistent)
- [ ]     The number of datasets on homepage equals total dataset amount and not open dataset amount
- [ ]     The number of datasets per on each Ministry on the ministry page is the total dataset amount and not open dataset amount
- [ ]     Navigating the website, there are no internal server errors
- [ ]     Clicking the search reset "x" when typing in a empty search box, resets the search box to empty

Pages to test:
- [ ] Tests should be done on individual ministry and group pages and the dataset search page

KatiRG:
issue clearing filters: if I select e.g. XLSX, it will display as checked and the filter pill will also display under the search bar. If I select XLSX again to clear it, it does not clear, and a second filter pill will be added:
Screenshot 2024-08-20 at 14 22 22

KatiRG:
The count in the filter pill sometimes show 0 instead of the count shown in the facet:


Screenshot 2024-08-20 at 14 30 43

aleeexgreeen:
> The count in the filter pill sometimes show 0 instead of the count shown in the facet:
> 
> Screenshot 2024-08-20 at 14 30 43

How can I reproduce this issue?

KatiRG:
> > The count in the filter pill sometimes show 0 instead of the count shown in the facet:
> > Screenshot 2024-08-20 at 14 30 43
> 
> How can I reproduce this issue?

This is reproducible on data4. Another example (on data4): choose ministry "Energy". The facet says "Energy (12)" but the pill shows "Energy (0)".

KatiRG:
> The number of datasets on homepage equals total dataset amount and not open dataset amount

On data4, there are 2849 datasets on the home page, whereas the sum of the access level counts on the dataset page is 2850. The extra count is due to one dataset that is private. To make it consistent, exclude `private = true`

KatiRG:
Some strange behaviour with pill counts again. Steps to reproduce (on data4):
1. go to datasets page and click `DIOXIN (2)` topic. Two open datasets will be displayed, as expected.
2. select `Forestry (3)` topic. Total 5 open datasets will be displayed, as expected.
3. now click `Restricted (1)` checkbox. Total 6 datasets will be displayed, as expected, however:
    - `DIOXIN (0)` pill shown with 0 counts
    - `DIOXIN` is no longer in the Topics facet list  

Screenshot 2024-08-22 at 13 20 44

A similar test:
1. go to datasets page and click `DIOXIN (2)` topic. Two open datasets will be displayed, as expected.
2. click checkbox `Restricted (0)`. Same two datasets are listed, as expected, however `DIOXIN (0)` shows 0 count.

aleeexgreeen:
> Some strange behaviour with pill counts again. Steps to reproduce (on data4):
> 
> 1. go to datasets page and click `DIOXIN (2)` topic. Two open datasets will be displayed, as expected.
> 2. select `Forestry (3)` topic. Total 5 open datasets will be displayed, as expected.
> 3. now click `Restricted (1)` checkbox. Total 6 datasets will be displayed, as expected, however:
>    
>    * `DIOXIN (0)` pill shown with 0 counts
>    * `DIOXIN` is no longer in the Topics facet list
> 
> A similar test:
> 
> 1. go to datasets page and click `DIOXIN (2)` topic. Two open datasets will be displayed, as expected.
> 2. click checkbox `Restricted (0)`. Same two datasets are listed, as expected, however `DIOXIN (0)` pill shows 0 count.

@KatiRG I was able to reproduce the issues described, but the issue is because of how the topics works. The topics return the 50 topics in descending order of dataset count. The last topic in displayed in the first test has a dataset count of 3, which is higher than 2 and that's why DIOXIN is not in the topic count. The issue is that topics are no longer reducing based on selected topics, as it does on prod. Maya suggested previously, not using the checkboxes on the topics until the topics/keywords are figured out, we may have to take this approach.

KatiRG:
Re low-count topics: As discussed with team, this case will resolve when the keywords are cleaned up, so we will let this pass for now.

ckanext/ontario_theme/plugin.py
```diff
@@ -1175,7 +1175,10 @@ def before_search(self, search_params):
         # pages
         elif fq:
             fq_list = search_params.setdefault('fq_list', [default_open])
-            facet_al = default_open if "access_level" not in fq else ""
+            if (fq.count("access_level") == 1 and 'access_level:"open"' in fq) or "access_level" not in fq:
```
KatiRG, line 1178: @aleeexgreeen  If "Restricted" is selected along with "Open", and a search term is entered in the search box, the results will still pertain to "Restricted" and "Open" datasets. Is this correct?
```
/dataset/?q=forestry&access_level=open&access_level=restricted&sort=score+desc%2C+metadata_modified+desc
```

aleeexgreeen: It would work in this instance as well, but to provide more context, to make the dataset search pages open by default the facet query (fq/fq_list) needs to have `{!bool tag=orFqaccess_level should=\'access_level:"open"\'}`. So both when opening a dataset page and when the form is submitted and only access level open is selected, the facet query needs to have that item. When there are more than one access_level selected the facet query needs to have `{!bool tag=orFqaccess_level should=\'access_level:"<access_level>"\' should=\'access_level:"<access_level>"\'}`, so this is to make sure upon opening a dataset search page the default is set to open, if the form is submitted on only open is selected, the results pertain to only open and if any facet is added, the results pertain to all facets selected

aleeexgreeen: I simplified this section further

ckanext/ontario_theme/fanstatic/scripts/odc-access-level.js
```diff
@@ -0,0 +1,70 @@
...
+            // Update checkboxes and callout
+            $(this.accessLevelBoxes).on('change', jQuery.proxy(this._onChange, this));
+            $(document).ready(jQuery.proxy(this._updateAccessLevelSentence, this));
+        },
+        _onChange: function () {
+            var form = $('form.search-form');
+            form.submit();
+        },
+        _updateAccessLevelSentence: function () {
+            var selectedBoxes = $('input[name="access_level"]:checked');
```
KatiRG, line 45: this is a `const`

ckanext/ontario_theme/fanstatic/scripts/odc-clear-filters.js
```diff
@@ -1,21 +1,27 @@
...
+ * CKAN module function for submitting form with empty facets once the clear
+ * filters button is clicked
+ */
+
+"use strict";
+
+ckan.module('clear_filters', function ($) {
+  return {
+      initialize: function () {
+          var facetsSelected = $(".filtered.pill");
```
KatiRG, line 15: this is a `const`

ckanext/ontario_theme/fanstatic/scripts/odc-clear-filters.js
```diff
@@ -1,21 +1,27 @@
...
+          if (facetsSelected.length >= 2) {
+              this.el.css({ display: "inline-block" });
+              this.el.on('click', jQuery.proxy(this._onClick, this));
+          }
+    },
+    _onClick: function (event) {
+        $(('#fields')).empty();
```
KatiRG, line 22: why double brackets?

KatiRG:
Can you djlint the html files?

ckanext/ontario_theme/plugin.py
```diff
@@ -1122,7 +1160,38 @@ def organization_facets(self, facets_dict, organization_type, package_type):
...
+        # To show all access levels on dataset counts on
+        # the homepage and org/group pages
+        if ('fq' not in search_params) or (fq and not fl):
+            fq_list = search_params.setdefault('fq_list', [])
+        # Show default open datasets or other access level on org/group search
+        # pages
+        elif fq:
+            fq_list = search_params.setdefault('fq_list', [default_open])
+            access_level_count = True if fq.count("access_level") > 1\
+                or (fq.count("access_level") == 1 and 'access_level:"open"' not in fq) else None
```
KatiRG, line 1179: `fq.count("access_level") > 1 or (fq.count("access_level") == 1` can be simplified to `fq.count("access_level") >= 1`

ckanext/ontario_theme/templates/internal/snippets/access_level_checkboxes.html
```diff
@@ -0,0 +1,33 @@
+{% set schema_dict = h.scheming_get_dataset_schema('dataset')['dataset_fields'] %}
+{% set name = "access_level" %}
+{%- asset "ontario_theme_scripts/ontario_theme_access_level_js" -%}
+{% with access_levels = h.scheming_field_by_name(schema_dict, name).choices %}
+  {% for levels in access_levels %}
+    {% set value = levels.get('value') %}
+    {% set count = h.get_facet_items_dict(name)|selectattr('name', '==', value)|map(attribute='count')|list %}
+    {% set display_name = h.scheming_choices_label(access_levels, value) %}
+    {% if checkboxes %}
```
KatiRG, line 9: It doesn't seem like `checkboxes` can ever be False

ckanext/ontario_theme/templates/internal/snippets/facet_arrangement.html
```diff
@@ -3,11 +3,12 @@
...
-{% set excluded_facets = [keywords] %}
-
+{% set excluded_facets = [keywords, "access_level"] %}
+{% set fields = h.scheming_get_dataset_schema('dataset')['dataset_fields'] %}
+ 
 {# Flag facets to be ignored from standard loop that calls facets. #}
 {% set single_option_facets = ["organization_jurisdiction", "organization_category"] %}
-
+{% do excluded_facets.append(single_option_facet) if single_option_facet else none %}
 {% if single_option_facets %}
```
KatiRG, line 12: No need to check `if single_option_facets` since they are hard-coded above in line 10

ckanext/ontario_theme/templates/internal/snippets/facet_arrangement.html
```diff
@@ -18,11 +19,10 @@
   {% endfor %}
 {% endif %}
 {% for facet in c.facet_titles %}
```
KatiRG, line 21: Only facets that are not in `excluded_facets` are actioned (line 28), so in line 21 you could already exclude `facet in c.facet_titles` that are `not in excluded_facets`.

Facets `organization_jurisdiction` and `organization_category` are in `excluded_facets`, so the `choices` variable need not be set i.e. they are never passed to the snippet.

excluded_facets: ['keywords_fr', 'access_level', 'organization', 'organization_jurisdiction', 'organization_category']

ckanext/ontario_theme/templates/internal/snippets/access_level_checkboxes.html
aleeexgreeen: `checkboxes` is defined when calling the snippet (line 78 of search_form.html) `{% snippet 'snippets/access_level_checkboxes.html', checkboxes=true %}` and (line 247 of search_form.html) `{% snippet 'snippets/access_level_checkboxes.html', extras=extras %}`. Checkboxes is always false for line 247

KatiRG:
@aleeexgreeen I found the following case where 1 dataset exists at level Restricted but when the Restricted access level is selected with the Open checkbox selected as well (also when all checkboxes are selected), that dataset is not shown and the callout says "No datasets":
Screenshot 2024-09-13 at 15 23 16

Steps to reproduce:
1. select Restricted then deselect Open
2. select Topic `aggregates (1)`
3. select Open checkbox (and can also select Under Review checkbox so that all checkboxes are selected)

ckanext/ontario_theme/templates/internal/snippets/facet_list.html
```diff
@@ -1,71 +1,89 @@
...
-        {% if (items|selectattr("active")|list|length) > 0 %}
-          {% set screen_reader_text = "Search filters are applied for "+title+"." %}
-        {% endif %}
-      {% endif %}
-    </h4>
-    <span class="sr-only">{{ screen_reader_text }}</span>
-  {% endwith %}
-{% endblock facet_list_heading %}
+{# use label_function to get translated titles of specific facets #}
+{% if name == "license_id" or name == "organization" or name == "groups" %}
```
KatiRG, line 7: can condense as `if name in ["license_id", "organization", "groups"]`

aleeexgreeen:
> @aleeexgreeen I found the following case where 1 dataset exists at level Restricted but when the Restricted access level is selected with the Open checkbox selected as well (also when all checkboxes are selected), that dataset is not shown and the callout says "No datasets": Screenshot 2024-09-13 at 15 23 16
> 
> Steps to reproduce:
> 
> 1. select Restricted then deselect Open
> 2. select Topic `aggregates (1)`
> 3. select Open checkbox (and can also select Under Review checkbox so that all checkboxes are selected)

Hi @KatiRG, I don't have the environment or access set up at the moment, but can you revert the change proposed in https://github.com/ongov/ckanext-ontario_theme/pull/598#discussion_r1750608059 and let me know if this is still an issue?

ckanext/ontario_theme/plugin.py
KatiRG: Thanks @aleeexgreeen ! This solves the problem. I didn't notice the brackets around the second `or` grouping! I'll revert

KatiRG:
> > @aleeexgreeen I found the following case where 1 dataset exists at level Restricted but when the Restricted access level is selected with the Open checkbox selected as well (also when all checkboxes are selected), that dataset is not shown and the callout says "No datasets": Screenshot 2024-09-13 at 15 23 16
> > Steps to reproduce:
> > 
> > 1. select Restricted then deselect Open
> > 2. select Topic `aggregates (1)`
> > 3. select Open checkbox (and can also select Under Review checkbox so that all checkboxes are selected)
> 
> Hi @KatiRG, I don't have the environment or access set up at the moment, but can you revert the change proposed in [#598 (comment)](https://github.com/ongov/ckanext-ontario_theme/pull/598#discussion_r1750608059) and let me know if this is still an issue?

Thanks @aleeexgreeen ! This solves the problem. I didn't notice the brackets around the second or grouping! I'll revert

ckanext/ontario_theme/plugin.py
```diff
@@ -1122,7 +1160,38 @@ def organization_facets(self, facets_dict, organization_type, package_type):
...
         '''
+        fl = search_params.setdefault("facet.field", [])
+        fq = search_params.get("fq", "")
+        default_open = '{!bool tag=orFqaccess_level should=\'access_level:"open"\'}'
+
+        ors = set(_get_default_ors())
+        # To show all access levels on dataset counts on
+        # the homepage and org/group pages
+        if ('fq' not in search_params) or (fq and not fl):
+            fq_list = search_params.setdefault('fq_list', [])
```
KatiRG, line 1173: `fq_list` is set here to `[]` but is never used again outside this `if` statement

ckanext/ontario_theme/templates/internal/snippets/search_form.html
```diff
@@ -47,166 +50,213 @@ <h2 class="ontario-label">
...
-                            <use xlink:href="#ontario-icon-close"></use>
-                          </svg>
-                        </span>
-                      </a>
-                      {%- if loop.index == 1 -%}</span>{%- endif -%}
+  <!-- block for package searches -->
+  {% block search_facets %}
+    {% if facets %}
+      <p class="filter-list">
+        {% if query %}
```
KatiRG, line 89: this is an empty `if` block

ckanext/ontario_theme/templates/internal/snippets/search_form.html
KatiRG, line 1: There are indentation irregularities in this file that are not caught by djlint for some reason. e.g. in block `{% block search_facets %}` and onwards.

ckanext/ontario_theme/templates/internal/snippets/search_form.html
```diff
@@ -47,166 +50,213 @@ <h2 class="ontario-label">
...
+              {% set callout_text = _("No datasets with access level") %}
+            {% endif %}
+            <div class="ontario-callout {{ border_highlight }} access-level-status ontario-small-12 ontario-large-7 ontario-columns">
+              <p id="access-level-sentence-id">
+                {{ callout_text.format(query=query|escape)|safe }}{{ _(':') }} <span id="access-level-sentence-value"></span>
+              </p>
+            </div>
+          {% endif %}
+          <div class="form-select form-group control-order-by search-page-select ontario-large-5 ontario-columns">
+            <label class="ontario-label" for="field-order-by">{{ _('Sort by') + _(':') }}</label>
```
KatiRG, line 211: When `count` = 0, the "Sort by' dropdown menu is still displayed on desktop screens, but on mobile the sort button is not displayed.

ckanext/ontario_theme/templates/internal/group/read.html
```diff
@@ -69,22 +63,20 @@
...
+               sol:category="primary"
+               viewBox="0 0 24 24"
+               preserveAspectRatio="xMidYMid meet">
             <use xlink:href="#ontario-icon-close"></use>
           </svg>
         </span>
-        <div>
-          {{ h.snippet('snippets/facet_arrangement.html', single_option_facets=single_option_facets, extras={'id':c.group_dict.id}) }}
-        </div>
+        <div>{{ h.snippet('snippets/facet_arrangement.html', extras={'id':c.group_dict.id}) }}</div>
```
KatiRG, line 79: In `organization/read.html`, you pass `single_option_facet="organization"` to `facet_arrangement.html`:

```
{{ h.snippet('snippets/facet_arrangement.html', extras={'id':c.group_dict.id}, 
single_option_facet="organization") }}
```
I'm wondering if you need to do the same for groups. Otherwise, all the groups will be listed in the facet column of an individual group page. 

Screenshot 2024-09-17 at 18 22 50
(screenshot from local)

ckanext/ontario_theme/fanstatic/scripts/odc-access-level.js
```diff
@@ -0,0 +1,70 @@
...
+
+ckan.module('access_level_checkboxes', function ($) {
+    return {
+        options: {
+            accessLevelBoxes: null,
+            openCheckbox: null,
+		},
+        initialize: function () {
+            this.accessLevelBoxes = $('input[name="access_level"][type="checkbox"]');
+            var params = new URLSearchParams(document.location.search);
```
KatiRG, line 24: `params` is a `const`

ckanext/ontario_theme/fanstatic/scripts/odc-clear-filters.js
```diff
@@ -1,21 +1,27 @@
...
+ * CKAN module function for submitting form with empty facets once the clear
+ * filters button is clicked
+ */
+
+"use strict";
+
+ckan.module('clear_filters', function ($) {
+  return {
+      initialize: function () {
+          const facetsSelected = $(".filtered.pill");
```
KatiRG, line 15: The code inside `initialize` is not necessary since the button will not be created in `search_form.html` (line 130) unless 2 or more facets have been selected.

ckanext/ontario_theme/templates/internal/snippets/search_form.html
```diff
@@ -47,166 +50,213 @@ <h2 class="ontario-label">
...
+                      </span>
+                    </a>
+                    {%- if loop.index == 1 -%}</span>{%- endif -%}
                     {% endfor %}
+                  </span>
                 </span>
-              </span>
+              {% endif %}
             {% endfor %}
             {% if c.fields|length >= 2 %}
```
KatiRG, line 130: Make it `>= 3` since there is always the `access_level` element in the `c.fields` array that will count as 1, e.g.:

Screenshot 2024-09-18 at 16 23 06

ckanext/ontario_theme/fanstatic/common.css
KatiRG, line 1: Change display to `inline-block`:

```
#clear-filters-button {
  display: inline-block;
}
```

and consolidate with line 643 which already defines css for `#clear-filters-button`.

ckanext/ontario_theme/plugin.py
aleeexgreeen: This is because the `fq_list` is used with `setdefault()`. The default value for fq_list is being set and modified within the if statement, so there is no need to use it outside the if statement, because it is being updated directly into `search_params`

ckanext/ontario_theme/templates/internal/snippets/search_form.html
aleeexgreeen: You may have a different version of djlint or what you are using is not specifically tailored for this repo. There are files which the command `pre-commit run --hook-stage manual format` works from (see `.djlintrc` and `djlint_rules.yaml`). I say this because when running `pre-commit run --hook-stage manual format` there are no formatting errors in the `search_form.html`

ckanext/ontario_theme/templates/internal/snippets/search_form.html
aleeexgreeen: This doesn't solve the problem, as you can have multiple access levels checked and `c.fields|length` will increase. All 3 access levels checked will have the same effect (The displaying of the button is all done in the javascript). I've opted to removing the if statement and leaving the javascript as if.

aimalkhan (approved):
👍

PR #605, opened 2024-10-08
Remove "Save and Add data" btn when creating new dataset

KatiRG:
## What this PR accomplishes
- Removes the "Save and Add Data" button when creating a new dataset, since that button leads into the old workflow that does not have validation checks.
- Renames the "Save and finish" button to "Publish Dataset"
- Changes content of newly-created, empty dataset page to read "This dataset is published with no data resources. Add a data resource to this dataset."

## What needs review
- Ensure that the "Save and Add Data" button is no longer there
- Ensure that the "Publish Dataset" button works
- Ensure that you can still edit an existing dataset and publish the updates


aimalkhan (approved):
changes look good.

PR #606, opened 2024-10-10
Remove hint text from data dictionary page

KatiRG:
## What this PR accomplishes
Removes hint text from data dictionary page under each column that said: "This can only be changed in the data file."

## What needs review
Confirm this text is no longer displayed on the data dictionary page.


aimalkhan (approved):
👍

PR #607, opened 2024-10-11
Update GA snippets and exclude specific paths from tracking

KatiRG:
## What this PR accomplishes
Updates GA snippets and excludes specific paths from being tracked by GA.

## What needs review
Review logic of the path matching. Only those paths that are not to be excluded should be passed to GA. 

Review that logic works for paths in test (current state) and test.data4 (state with validation feature).


ckanext/ontario_theme/templates/internal/gtm.html
```diff
@@ -1,4 +1,4 @@
 <!-- Google Tag Manager -->
-<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start': new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src= 'https://www.googletagmanager.com/gtm.js?id='+i+dl+ '&gtm_auth=Wq24uXP3CoptjFHA01X7vA&gtm_preview=env-26&gtm_cookies_win=x';f.parentNode.insertBefore(j,f); })(window,document,'script','dataLayer','GTM-TK6PW9Z');</script>
+<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+'https://www.googletagmanager.com/gtm.jsid='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','GTM-TK6PW9Z');</script>
```
aimalkhan, line 3: is `jsid` correct? the snipped shared with us has `js?id` instead

aimalkhan (commented):
just wondering if the gtm.html file is correct

aimalkhan (approved):
👍  works locally; pushing to test to see changes.

PR #615, opened 2024-11-26
Patch: keep invalid resources in FileStore

KatiRG:
## What this PR accomplishes
- Keeps resources that fail validation in FileStore so that they can still be downloaded from the displayed URL 
- Displays the "Access data" block to allow download from FileStore instead of from DataStore to keep consistency
- Displays a small explanatory text that the file is in the process of being fixed

## What needs review
Check that the behvavious is as expected for:
- editing an existing valid resource with an invalid one
- editing an existing valid resource with a valid one
- creating a new valid resource
- creating a new invalid CSV resource
- creating a new non-CSV resource

Note: this PR must be checked with the accompanying changes in ckanext-validation (https://github.com/ongov/ckanext-validation/pull/11).


aimalkhan (approved):
looks good!

PR #618, opened 2024-12-12
Show download box and API box for CSV files in DataStore prior to validation launch

KatiRG:
## What this PR accomplishes
Modifies logic for showing the download box + API box for resources that exist in the DataStore before the launch of the validation feature (i.e. they have no validation badge and no `validation` field in the `resource` object).

## What needs review
Verify that all cases show content in the right hand side of the resource page as expected:

- Resource in DataStore before the launch of the validation feature (ie there is no badge): Should show Download box with file types, and the API box
- Resource not in DataStore: Should show “Access data” in the right hand box and link to the file
- Resource in DataStore but validation is False: Should show “Access data” in the right hand box and link to the file
- Resource in DataStore and validation is True: Should show Download box with file types, and the API box
- AI resource: Should show “Access AI” in the right hand box and link to the COVID site
- non-CSV resources: check any PDF, DOCX, etc.: Should show “Access data” in the right hand box and link to the file

### DIFF
--- a/ckanext/ontario_theme/templates/internal/scheming/package/resource_read.html
+++ b/ckanext/ontario_theme/templates/internal/scheming/package/resource_read.html
@@ -52,7 +52,9 @@
           {% block resource_read_title %}
             <h1 {% if "https://" in h.resource_display_name(res) -%}
                 class="break-all"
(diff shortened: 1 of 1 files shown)


aimalkhan (approved):
👍

PR #619, opened 2024-12-18
[Task] Allow formats for remotes to be changed manually

KatiRG:
## What this PR accomplishes
Reverts to original code where resource format can be manually entered and removes the automatic application of the `WEB` label to remotes where no format is manually specified. 

Notes: 
- remotely-hosted CSV files will be run through the validation process, but the resource page will not show the Download box or the API box. Since the remote host is the most up-to-date, the catalogue database could potentially diverge significantly from it. 
- the content for the Data Dictionary page created for the case of remote CSV files has been removed, since the validation process is no longer being skipped for remote CSVs.
- manually entered formats that contain a typo (e.g. CVS) or are not in the accepted format list (e.g. MYFORMAT) will not be checked or corrected and will pass as is.

## Issue(s) addressed
DATA-1630

## What needs review
Must be tested with https://github.com/ongov/ckanext-validation/pull/14 branch checked out.

To test, link to remote files of various formats (ZIP, CSV, etc) and manually enter the format. Options:

-   enter a format that is the same as the format of the remote resource
-   enter a format that is different from the format of the remote resource
-   enter a format type that has a typo or make up a format (e.g. MYCSV, BLANK, etc)
-   leave the format blank

Also check that the regular sanity tests also pass (i.e. locally-uploaded CSV files both valid and corrected, and non-CSV files, work as expected.)

### DIFF
--- a/ckanext/ontario_theme/resource.py
+++ b/ckanext/ontario_theme/resource.py
@@ -19,6 +19,9 @@ import ckan.lib.base as base
 import ckan.logic as logic
 import ckan.lib.helpers as h
 
+import logging
+log = logging.getLogger(__name__)
+
 NotFound = logic.NotFound
 NotAuthorized = logic.NotAuthorized
 ValidationError = logic.ValidationError
@@ -146,6 +149,7 @@ class CreateView(MethodView):
     def get(
         self, package_type, id, data=None, errors=None, error_summary=None
     ):
+        print('HEJ resourcepy def get of CreateView')
         # get resources for sidebar
         context = {
             u'model': model,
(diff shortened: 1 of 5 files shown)


ckanext/ontario_theme/resource.py
```diff
@@ -19,6 +19,9 @@
 import ckan.logic as logic
 import ckan.lib.helpers as h
 
+import logging
+log = logging.getLogger(__name__)
```
aimalkhan, line 23: we can remove the logging code.

aimalkhan (changes requested):
let's remove the unused code and we can go ahead!

aimalkhan (approved):
looks good 👍

PR #620, opened 2024-12-20
rm check for validation_status in resource

KatiRG:
## What this PR accomplishes
Removes a check for `validation_status` in the `resource` object which prevents pre-validation resources from being updated.

## What needs review
Needs to work for all cases (pre-validation resources, post-validation resources, new resources).


aimalkhan (approved):
👍

PR #623, opened 2025-02-11
Restore UI dictionary table on resource page

KatiRG:
## What this PR accomplishes
Restores the UI block to display the data dictionary table in the resource page of CSV resources, styled according to the Ontario Design System.

## Issue(s) addressed
DATA-1006

## What needs review
Verify that the data dictionary table:
- is displayed on the resource page of locally-uploaded CSV files (and not remote CSV files)
- is styled according to the Ontario Design System (see ticket)
- shows three columns: Column, Type, and Description
- has capitalized data types
- works on mobile


KobikahChandran:
@KatiRG if the file is labelled as EN and FR, I don't think the colour scheme is working for the table. The body of the table is only in the white shade.

KobikahChandran (approved):
Perfect!

PR #625, opened 2025-05-13
set ignore hash to False to ignore unchanged files

KatiRG:
## What this PR accomplishes
Ignores resources whose hash has not changed, rather than always pushing to datastore even when the data file has not changed.

## Issue(s) addressed
DATA1668

## What needs review
A full test of validation test cases to ensure that data is still being properly uploaded when data edits are made and when metadata edits only are made.


KatiRG:
Update: we actually need to ignore the hash to fix existing resources that are not ordered starting at `1` in the `_id` column. Retracting PR.

PR #626, opened 2025-07-30
Redirect to publish page when dictionary form left blank

KatiRG:
## What this PR accomplishes
Resolves the 500 server error when the data dictionary form is left completely blank by redirecting to publish page after Step 3. 

Also takes care of case when the data type field not filled in but the description field is added. In this case, the modified data dictionary needs to be pushed to database, same as when the data type is defined.

## Issue(s) addressed
DATA-1678

## What needs review
Test
- creating a new resource without filling out the data dictionary form redirects to publish page without error
- creating a new resource and filling out description field only (leaving data type untouched) in data dictionary form redirects to publish page without error and the description is listed in the publish page
- creating or editing a resource with data dictionary form filled out works as normal

Also test that


aimalkhan (commented):
change looks good, and will merge after the pipeline fix

PR #629, opened 2025-08-27
remove AI-related content

KatiRG:
## What this PR accomplishes
Removes AI-related content from the code base. This includes references to "algorithms", which is also obsolete.

## Issue(s) addressed
DATA-1692

## What needs review
Check that there are no longer any instances of AI or algorithms in the code base, and that there is no impact on the site.


aimalkhan (approved):
references to "algorithm" in `internal/..assets.less` will be resolved with other tech debt

PR #635, opened 2025-12-24
replace legacy ._data in license obj and ensure get_license returns a dict

KatiRG:
## What this PR accomplishes
CKAN 2.11 replaced license objects with dictionaries, so any calling of `._data` will throw an error. This PR replaces the legacy sytnax in the entire code base and adds a check to `get_license()` to ensure that a dictionary is returned, even in the case of empty license.

These fixes allow a dataset page or resource page with no license to load.

## Issue(s) addressed
DATA-1733, DATA-1729 (seems that PR #631 did not fix the issue)

## What needs review
-   verify that a dataset and a resource with no license can be accessed from home page (Dataset page is currently not working so don't try from there)
- verify that datasets and resources with a license still work
-   test EN and FR

### DIFF
--- a/ckanext/ontario_theme/plugin.py
+++ b/ckanext/ontario_theme/plugin.py
@@ -654,7 +654,30 @@ def get_recently_updated_datasets():
 def get_license(license_id):
     '''Helper to return license based on id.
     '''
-    return Package.get_license_register().get(license_id)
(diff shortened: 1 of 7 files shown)


aimalkhan (approved):
looks good!

PR #638, opened 2026-01-12
handle empty dataset metadata fields

KatiRG:
## What this PR accomplishes
Makes the code more robust by handling empty dataset metadata fields in the fluent schema (ie that have _translated counterparts).

## Issue(s) addressed
DATA-1759

## What needs review
I found this error when trying to create a new resource in a dataset page with missing metadata fields. To test, create a new resource in a dataset page that is missing some metadata fields in the fluent schema (ie fields that have _translated counterparts)


aimalkhan (approved):
looks good! can't reproduce the issue with this change.

PR #640, opened 2026-01-28
Add WEB as Resource File Type Option for Remote Links

farazkabbo:
### What this PR accomplishes

- Added `WEB` as an additional option in the **Resource File Type** dropdown for remote links.
- Updated `get_accepted_format_options()` in `plugin.py` to dynamically append `WEB` at runtime.
- Since `WEB` is not a MIME type or standard file type, it is **not** included in `accepted_resource_formats.json`. Instead, it is appended dynamically in the plugin to allow users to select `WEB` when adding resources that are remote URLs.
- Ensured that standard MIME type validation logic remains unaffected by this addition.

### Issue(s) Addressed

- DATA-1765

### What Needs Review

- Verify that `WEB` appears in the **Format** dropdown when adding or editing a resource.
- Select `WEB` for a resource with a remote URL and confirm that:
  - The selection saves correctly.
  - The selected format displays as expected after saving.
- Verify that other file format options continue to function normally.
- Confirm that MIME type validation behavior is not affected by the dynamic addition of `WEB`.

### DIFF
--- a/ckanext/ontario_theme/plugin.py
+++ b/ckanext/ontario_theme/plugin.py
@@ -648,6 +648,36 @@ def get_license(license_id):
     return Package.get_license_register().get(license_id)
 
 
+def get_accepted_format_options():
+    '''Helper to return accepted resource format options for select dropdown.
+    Reads from accepted_resource_formats.json and returns a list of dicts
+    with 'value' (format extension) and 'text' (description) keys.
+    '''
+    import os
+    resource_format_path = os.path.join(os.path.dirname(__file__),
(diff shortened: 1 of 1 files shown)


aimalkhan:
Let's 
1. improve the title of the PR
2. mention the ticket DATA-* it is solving like [this ticket](https://github.com/ongov/ckanext-ontario_theme/pull/635) does

aimalkhan:
@farazkabbo - can you fix the issue with the PR title?

farazkabbo:
Done @aimalkhan

PR #649, opened 2026-06-09
DATA-1460: Design Update of the History (Activity Stream) Page

AbiJ17:
## What this PR accomplishes
- Updated dataset activity history UI to match approved mockups
- Removed activity icons
- Added spacing/alignment adjustments between dates and activity text
- Added admin-only visibility for OPS employee names
- Added clickable profile links for admin users
- Ensured public users cannot see OPS employee names
- Added approved French translation for: “dataset was updated”

## Issue(s) addressed
- DATA-1460

## What needs to be reviewed
- Verify activity history formatting matches mockups
- Confirm dataset links work correctly
- Confirm OPS employee names are hidden from public users
- Confirm admin users can see clickable profile links
- Verify approved French translation displays correctly

Screenshot 2026-06-09 133205
Screenshot 2026-06-09 132951
Screenshot 2026-06-09 132327


ckanext/ontario_theme/templates/internal/snippets/activity_stream.html
```diff
@@ -0,0 +1,36 @@
...
+
+      {% set dataset_link = h.link_to(
+          activity.data.get('package', {}).get('title', ''),
+          h.url_for('dataset.read', id=activity.data.get('package', {}).get('name', ''))
+      ) %}
+
+      {% if activity.activity_type == 'new package' %}
+          {{ dataset_link }} dataset was created
+      {% elif activity.activity_type == 'changed package' %}
+          {% if h.lang() == 'fr' %}
```
aimalkhan: use .po file for fr translations

aimalkhan (changes requested):
let's fix the translation issue.

ckanext/ontario_theme/templates/internal/snippets/activity_stream.html
AbiJ17: Added .po file workflow for fr translations. I also recompiled the .mo file to include the updated translations, and verified that the translations show up on the page as expected.

ckanext/ontario_theme/i18n/fr/LC_MESSAGES/ckanext-ontario_theme.po
```diff
@@ -832,1459 +1035,3004 @@ msgstr "\n"
...
+msgstr ""
+
+#: ckanext/ontario_theme/templates/internal/user/new_user_form.html:18
+msgid "Create Account"
+msgstr ""
+
+#~ msgid "Queen&#8217;s Printer for Ontario"
+#~ msgstr "Imprimeur de la Reine pour l’Ontario"
+
+#~ msgid "Contact Us"
```
aimalkhan: where are these lines that are commented out coming from?

ckanext/ontario_theme/i18n/fr/LC_MESSAGES/ckanext-ontario_theme.po
```diff
@@ -832,1459 +1035,3004 @@ msgstr "\n"
...
+
+#: ckanext/ontario_theme/templates/internal/user/new_user_form.html:11
+msgid "Confirm"
+msgstr ""
+
+#: ckanext/ontario_theme/templates/internal/user/new_user_form.html:18
+msgid "Create Account"
+msgstr ""
+
+#~ msgid "Queen&#8217;s Printer for Ontario"
```
aimalkhan: the "Queen's printer" is not currently being used but it is a valid translation, and we might need to use it in the future. So let's keep it.

ckanext/ontario_theme/i18n/fr/LC_MESSAGES/ckanext-ontario_theme.po
aimalkhan: the PR here: https://github.com/ongov/ckanext-ontario_theme/pull/547/changes does a similar addition to what we want but it does not need an entire reformat of the page. Can you try to preserve the original formatting other wise it will be really hard to review what was changes and what wasn't

aimalkhan (changes requested):
Can you please check why the format changed and extra commented lines were added to the .po file? It is pretty hard to review the code like that.

ckanext/ontario_theme/i18n/fr/LC_MESSAGES/ckanext-ontario_theme.po
AbiJ17: Reverted back to original version of .po file, added French translation for dataset was updated at the bottom of the file.

ckanext/ontario_theme/i18n/fr/LC_MESSAGES/ckanext-ontario_theme.po
AbiJ17: Seems like this is already included in the original .po file.

ckanext/ontario_theme/i18n/fr/LC_MESSAGES/ckanext-ontario_theme.po
```diff
@@ -2022,10 +2022,6 @@ msgstr "Au fur et à mesure que nous travaillons à son amélioration le catalog
 msgid "<a href=\"%(url)s\">More %(group_title)s</a>"
 msgstr "<a href=\"%(url)s\">Plus %(group_title)s</a>"
 
-#: ckanext/ontario_theme/templates/internal/header.html:89
-msgid "Ministries"
```
aimalkhan, line 2026: do we not need the Ministries translation?

AbiJ17: When I tried to recompile the .mo file after reverting the .po file to its original version, I got a duplicate definition error and it pointed to lines 101 and 2026. There is already a Ministries translation at Line 101, so I removed the one at 2026 and kept the one at 101, after which it was able to recompile. The translation seems to show up on my end.

Screenshot 2026-06-12 103632
Screenshot 2026-06-12 104224

AbiJ17:
> Can you please check why the format changed and extra commented lines were added to the .po file? It is pretty hard to review the code like that.

Reverted the .po file back to what it was originally, and added the new translations to the bottom of the .po file.

aimalkhan (approved):
looks good!

KatiRG:
Adding a follow-up review note for transparency after a deeper technical review. I recommend reverting this merge and related PRs #652 and #659, then continuing implementation from the [DATA-1460_activity_stream branch](https://github.com/ongov/ckanext-ontario_theme/tree/DATA-1460_activity_stream):

Key issues identified:

1. Determination of activity type overrides native CKAN behaviour and replaces with custom, hard-coded if/else logic, which also misses some activity types such as when a new group is created.
2. Retrieval of the dataset name uses raw stored values instead of CKAN helpers that provide a more reliable current display (e.g. in case display name has become stale).

PR #658, opened 2026-07-22
DATA-1519: Mandatory Metadata Fields

AbiJ17:
## What this PR accomplishes

- Makes File, Format, Name (EN and FR), Description (EN and FR), Type, Language, and Data Last Updated mandatory on the Add/Edit Resource page.
- Adds required-field indicators and validates all mandatory fields when **Continue** is clicked.
- Displays a validation summary and inline errors for missing fields.
- Allows either a URL or uploaded file to satisfy the resource requirement.
- Prevents missing translated values from causing an internal server error.

## Issue(s) addressed

- DATA-1519
- Adds onto DATA-1069 & [PR 590](https://github.com/ongov/ckanext-ontario_theme/pull/590)

## What needs to be reviewed

- Verify all listed fields are mandatory when adding or editing a resource.
- Verify missing fields appear in the validation summary and display inline errors.
- Verify URL-only and file-only resources can both be submitted successfully.
- Verify leaving both URL and File blank prevents submission.
- Verify completing all mandatory fields allows the workflow to continue.


ckanext/ontario_theme/templates/internal/scheming/form_snippets/upload.html
```diff
@@ -0,0 +1,20 @@
+{% import 'macros/form.html' as form %}
+
+{%- set is_upload = (data.url_type == 'upload') -%}
+{{ form.image_upload(
+    data,
+    errors,
+    field_url=field.field_name,
+    field_upload=field.upload_field,
+    field_clear=field.upload_clear,
+    is_upload_enabled=h.uploads_enabled("resource") if h.check_ckan_version(min_version="2.12", max_version="3.0") else h.uploads_enabled(),
```
aimalkhan: what is this version check for ckan? we are still on ckan 2.9, so seeing 2.12 is a bit of a surprise

AbiJ17: Good catch, I was having an issue with the page navigation as after I clicked Continue on the Add or Edit Resource page after entering the URL/file and all the mandatory fields, it kept reloading the same page instead of moving to the next page. I didn't realize the suggestion was intended for CKAN 2.12, this was my fault. I have updated the code to call h.uploads_enabled() directly, which works in CKAN 2.9.

